PR #3, opened 2014-10-01
Initial backend commit

cmbrose:
This is a proof of concept for backend functionality.  The example can create/stop/remove a test container through REST handles on your local machine.

To use you will need Docker and the [Go AppEngine SDK](https://cloud.google.com/appengine/downloads) on your machine.
1. Navigate into the /docker-manager/backend folder
2. Run `$> path/to/go/SDK/dev_appserver.py gotest`
3. Opening localhost:8080 in a browser will display some basic info about your images/containers/etc
4. Hitting /create will create a container named `test_container` on your machine
   - **NOTE**: If you don't have the image `training/webapp:latest` this hit will probably take multiple minutes the first time (as your machine pulls it down).  If you don't see an explicit error just be patient :)
5. Hitting /stop will stop `test_container`
6. Hitting /remove will remove `test_container` from your machine


cmbrose:
@ngmiller @chfogerty @theMagicalKarp @zach-taylor

zach-taylor:
Hey @cmbrose, this looks great so far.

Would it be possible to use a web framework for the backend? I know this is a PoC, but it could save us some time and effort. Plus, some MVC action would be nice.

A quick Google search netted me these: http://martini.codegangsta.io/ and https://revel.github.io/

nickmonad:
:shipit: Looks good! Looking forward to going over it more on Friday :)

theMagicalKarp:
Looks awesome, however I'm not sure if this works on app engine.  I know it works locally but as soon as you try to deploy it, there might be a few problems.  From what I can tell we will have to use something like https://cloud.google.com/appengine/docs/go/urlfetch/ to make any sort of request from inside the app.

Also since we are on the topic Go web frameworks I think [gin](http://gin-gonic.github.io/gin/) is worth mention.  Martini is awesome as well.

cmbrose:
@zach-taylor @theMagicalKarp I'm 100% open to framework discussions in the near future.  Like Zach mentioned, this was more PoC of the go-dockerclient library's capabilities.

@theMagicalKarp Thanks for the tip! I'll do some looking into that tomorrow.

@ngmiller That's a good call.  I've added them.

backend/gotest/test.go
```diff
@@ -0,0 +1,211 @@
+package main
+
+import (
+    "fmt"
+    "net/http"
+    "github.com/fsouza/go-dockerclient"
```
nickmonad, line 6: wat

theMagicalKarp, line 6: Yeah, go package management is pretty nice.
nod

cmbrose:
> From what I can tell we will have to use something like https://cloud.google.com/appengine/docs/go/urlfetch/ to make any sort of request from inside the app.

@theMagicalKarp if this is something we end up needing to use, we will need to address it soon.  I make no direct requests to the client's Docker, that is all handled through the go-dockerclient library.  Specifically, the library sets up the http.Client instance that your link references in [client.go](https://github.com/fsouza/go-dockerclient/blob/master/client.go).

However, the library just uses a default connection

```
return &Client{
        HTTPClient:          http.DefaultClient,
        // ...
    }
```

If http.DefaultClient does not work, we could potentially override that field with our own http.Client that we make using the URL Fetch example.  If the override doesn't work, then the go-dockerclient library may not be very useful to us...

theMagicalKarp:
@cmbrose I agree. 

Have you seen this yet https://github.com/fsouza/go-dockerclient/issues/123? This is actually where I pulled the code for setting up a TLS connection.  If that stuff doesn't work, I don't think there is much hope either.

This is why I'm hesitant on targeting app engine as our platform.  If we interface with docker we might have to redo work others have already accomplished.

Otherwise I'm all for deploying to a hosting provider instead of app engine.  If we dockerize our app things should be pretty streamlined as far as deploying is concerned.

Just this morning I setup a coreos instance on digitial ocean, pulled the go implementation of [docker-thing/test](https://github.com/theMagicalKarp/docker-thing/tree/test) off of [themagicalkarp/docker-thing:test](https://registry.hub.docker.com/u/themagicalkarp/docker-thing/), setup a certificate, built server/client keys, pointed one of my subdomains to the instance, and ran.  Once everything is configured it's a simple as calling `docker pull themagicalkarp/docker-thing` from the server to update our app when ever we have changes.

It's currently running here at http://dev.rob-sheehy.com.  I'll also email you guys the keys, so that you can access that  docker via go-dockerclient or generic docker client.

chfogerty:
The project description mentions GCE as a potential target host. If we're fighting with GAE for our first release, just to rewrite it all later, we may as well just go for GCE now.

theMagicalKarp:
Yay! ◕ ◡ ◕

nickmonad:
For the sake of communication, I'm all for deploying elsewhere. Ultimately, it would be nice to have this be a standalone application that could potentially be hosted anywhere. :+1:

cmbrose:
There is a lot of mis-communication going on about the intentions of this request so hopefully I can clear some of that up.

This is purely intended to test the capabilities of the go-dockerclient library.  I have zero intentions of this becoming a fully-fledged backend application (and it's not written at all in a way that would make doing that be a pleasant experience).

I used App Engine only because local deploys on it are incredibly easy.  I do not intend for my use of App Engine here to cause us to use it as out backend host.

Hopefully this helps clear things up.  I'm sorry for all the confusion that it has caused.

nickmonad:
@cmbrose No worries at all! Frankly, we're probably talking about it too much up front anyway. We'll certainly iterate as the requirements/stability changes. For the record, my comment was not meant to be "trash this until we get GCE or some other platform working" - in fact, if this works locally, I'd say it's good for a merge. Otherwise, we can wait to merge the initial back end code in tomorrow evening if getting GCE/Digital Ocean up and running is as simple as it looks.

+1

theMagicalKarp:
Agreed +1

zach-taylor:
:+1:

PR #4, opened 2014-10-07
Init Dockerfile to Dockerize lighthouse

theMagicalKarp:
## What

Here's a simple Dockerfile we can use to build a lighthouse image locally and push up to our private registry, "reg.rob-sheehy.com".
## Setting up
- Add the lighthouse repo to your "shared folders" for boot2docker inside virtual box.
  screen shot 2014-10-07 at 11 12 54 am
- Restart boot2docker `boot2docker restart`
## Running build
- cd to the root of lighthouse, i.e. wherever the Dockerfile is
- run `docker build -t reg.rob-sheehy.com/some-name-for-image .`
## Workflow

This is what I'm imagining our workflow could be.  This by no means is set in stone, I'm just trying to demonstrate how Dockerizing our app can be beneficial.
- Write Code
- Push code changes to new branch on the lighthouse repo
- Build Image locally with Dockerfile
- Push Image to private registry 
- QA Pulls Images from private registry 
- QA verifies changes by running image
- QA merges branch into master
- Someone pulls image into production and runs it
## Notes

Feel free to pull the test image I pushed up to our registry that I built locally with this Dockerifle.  `docker pull reg.rob-sheehy.com/lighthouse`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,66 @@
+# Ubuntu with python & node
+FROM dockerfile/nodejs
+
+MAINTAINER lighthouse
+
+
+# ----- Install go -----
+WORKDIR /root
+RUN \
+  mkdir -p /goroot && \
+  curl https://storage.googleapis.com/golang/go1.3.1.linux-amd64.tar.gz | tar xvzf - -C /goroot --strip-components=1
+
+ENV GOROOT /goroot
+ENV GOPATH /gopath
+ENV PATH $GOROOT/bin:$GOPATH/bin:$PATH
+
+WORKDIR /data
+
+
+# ----- Add private ssh key -----
+RUN mkdir -p /root/.ssh
+
+# Private deploy key for lighthouse, "It's a secret to everybody."
+RUN echo "-----BEGIN RSA PRIVATE KEY-----\n\
+MIIEowIBAAKCAQEAu5RKR50Jr362XY5/QukNsw2r6rSDQLS5EZxlitXPqprDEC4w\n\
+nt9GIC3mbnUa+RvG0Sx2p3ZSLyt8rh4W8Emajkyj77V3Ycoc35L13Z5sda9EUX0z\n\
+Y08UkXkK58Kihjs3GgbLH1NaUlmVqI8QHDmIKsJi2Kzp5ufja8pZII4ifVZWWmhu\n\
+Yt1JzekxhQNSDY5NDMPhV7NKdz6qMnMgvJtBsORbcharvtgvcqOliP1pleQs5zzi\n\
+MYnJeUu+W8HFoqjNLIdxX9CK/weXN8LZLxlE1QsJ2OI1uxEXTthtf1LyB61U2BAR\n\
+xpZM4imZR8mIx+SPwNwbMxMhCQz+tIALRYMxRwIDAQABAoIBACQtlaX6Q8P1THb+\n\
+5Myi5mGCYYYDCs2QDaG36F2+ny7oanbUccwyg/Pw5mCndWxWTyJI0Rm7WF6ApKtw\n\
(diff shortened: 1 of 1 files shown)


theMagicalKarp:
@ngmiller @chfogerty @cmbrose @zach-taylor Let me know your thoughts/ideas or if anything isn't working for you.

Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+ENV PATH $GOROOT/bin:$GOPATH/bin:$PATH
+
+WORKDIR /data
+
+
+# ----- Add private ssh key -----
+RUN mkdir -p /root/.ssh
+
+# Private deploy key for lighthouse, "It's a secret to everybody."
+RUN echo "-----BEGIN RSA PRIVATE KEY-----\n\
```
nickmonad, line 24: :fearful:

nickmonad, line 24: Oh, we can remove this once the repo is public right?

theMagicalKarp, line 24: Yeah I know it's a little scary but I'm not worried about it.
- Our private registry is "private".
- Our private github is "private".
- Once this repo goes public all we need to do is remove the new public key from the list of deploy keys on our repo.

theMagicalKarp, line 24: We will just have to retire both the private and public keys, since this will show up in our git history.

theMagicalKarp, line 24: danger zone

nickmonad:
+1

cmbrose:
+1

Dockerfile
```diff
@@ -0,0 +1,66 @@
+# Ubuntu with python & node
+FROM dockerfile/nodejs
```
zach-taylor, line 2: Is there a reason we are making our base on the nodejs image rather than the golang image?

There's an official golang image here: https://registry.hub.docker.com/_/golang/

theMagicalKarp, line 2: I felt this was a good idea because we are probably going to use a node for building the angular stuff & using stuff like Gulp.
However, it might be a good idea to build from the [Go](https://registry.hub.docker.com/u/dockerfile/go/dockerfile/) image instead of [node](https://registry.hub.docker.com/u/dockerfile/nodejs/dockerfile/) image because node has the extra dependency of python.  I'm not sure why they did that. :/  That means instead of manually installing go we should manually build node.
It might be worth a shot though.

theMagicalKarp, line 2: Alright, for what ever reason running [dockerfile/node](https://registry.hub.docker.com/u/dockerfile/nodejs/dockerfile/) without python, fails. :/

Also I think the diference between [dockerfile/nodejs](https://registry.hub.docker.com/u/dockerfile/nodejs/dockerfile/) vs [node](https://registry.hub.docker.com/_/node/) is the first one is for building and running exclusively inside a container while the later is a pure node shell used for running cross-container commands, such that the container doesn't have to have node.  I guess I'm still not exactly sure about the later yet. 

Personally I would rather stick to the dockerfiles because they are simpler to use and more customizable, even though they are a little more inefficient.

As for switching to [dockerfile/go](https://registry.hub.docker.com/u/dockerfile/go/) I don't think that would be the best idea because we will have to manually install python & node vs just manually installing go.

Thoughts?

nickmonad, line 2: I agree. Installing node inside the Dockerfile looks a bit hairy. It's easy to follow what's going on, but that means less for us to maintain long term if that install process breaks for some reason.

The same guy notes on the `dockerfile/nodejs` and `node` comment threads that the node image is approx. 200 MB larger. For now, what we have seems to be the easiest, simplest solution. It would be interesting to experiment and compare image sizes though. Especially if we based off ubuntu/coreos and installed everything manually.

zach-taylor, line 2: Yeah, it looks like something in the Node install process uses Python. I also think what we have right now will work fine.

We can look into sizes/build times later on, if we have time.

+1

PR #5, opened 2014-10-09
The Great Front-end Migration

nickmonad:
https://trello.com/c/llYodwbl/29-move-front-end-from-docker-thing-to-docker-manager

The development bootstrapping process is (maybe) complete. How to run these changes:
1. fire up a boot2docker VM and set up the shared directory as specified in Rob's previous PR
2. open up Dockerfile and uncomment the branch switch to be: `RUN git checkout front-end`
3. `docker build -t test/some-tag .`
4. `docker run -t -i -p 5000:5000 test/some-tag python -m SimpleHTTPServer 5000`

Once the backend server is setup to serve the initial request for `index.html`, we can remove the WORKDIR modifications and that python command for the container run.

Let me know if anything seems odd. Thanks!

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -1,0 +1,72 @@
+<!DOCTYPE html>
+<html lang="en" ng-app="dockerThing">
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
+    <title>Docker Thang</title>
+
+    <link rel="stylesheet" href="/static/css/picnic.css">
+    <link rel="stylesheet" href="/static/css/base.css">
+</head>
+
+<body>
+
+<nav>
+  <a href="#" class="main" style="">Docker Thing</a>
+</nav>
+
+<div class="container" ng-controller="mainController">
+    <h6>Attention all planets of the solar federation... we have assumed control.</h6>
+
+    <h1>Containers</h1>
+    <div id="containers" class="row">
+        <table class="primary">
+            <thead>
+                <tr>
+                    <th>Image</th>
+                    <th>Status</th>
+                    <th>Created</th>
+                    <th>Ports</th>
+                    <th>Command</th>
+                    <th>Names</th>
+                </tr>
+            </thead>
+            <tbody>
+                <tr ng-repeat="container in remote.containers">
+                    <td>{{ container.Image }}</td>
+                    <td>{{ container.Status }}</td>
+                    <td>{{ formatTime(container.Created) }}</td>
+                    <td>{{ container.Ports[0].IP }} : {{ container.Ports[0].PublicPort }} -> {{ container.Ports[0].PrivatePort }} / {{ container.Ports[0].Type }}</td>
+                    <td>{{ container.Command }}</td>
+                    <td>{{ container.Names }}</td>
+                </tr>
+            </tbody>
+        </table>
+    </div>
+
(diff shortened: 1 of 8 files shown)


nickmonad:
@cmbrose @chfogerty @theMagicalKarp @zach-taylor 
For your review

cmbrose:
Works for me! +1

client/app/index.html
```diff
@@ -0,0 +1,72 @@
+<!DOCTYPE html>
+<html lang="en" ng-app="dockerThing">
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
+    <title>Docker Thang</title>
+
+    <link rel="stylesheet" href="/static/css/picnic.css">
```
theMagicalKarp, line 8: We can probably rely on any css frameworks to be on an external source/cdn. This would be to our advantage since,
- we don't have to worry about hosting it
- it's fast and more likely to be cached on a user's client

No big deal though I realize it's not an immediate, just food for thought as we move forward.
On that note I think we should start looking into what type css framework we want to use. 
In my opinion....
- [Picnic CSS](http://picnicss.com/) is sleek and sexy, but kinda limited.
- [Bootstrap](http://getbootstrap.com) is good looking, has great documentation, and very versatile.  However I feel it's been overused by student projects and is a little bloated. It's 113kb not mention the extra js you need to load to get complete functionality.
- [skeleton](http://www.getskeleton.com/) is "bare bones", and their responsive grid is my favorite. 
- [purecss](http://purecss.io/) is also pretty nice, it's kinda similar to picnic.

I'm a fan of any of these guys and I'm open to learning about new ones.

Also for anyone interested on modern web design, google has an awesome site for referencing stuff like this. https://developers.google.com/web/fundamentals/

theMagicalKarp:
+1 looks good

client/app/index.html
nickmonad, line 8: Sounds good. I'll keep the external sources in mind for CSS. As for the framework, bootstrap is hard to beat IMHO :) Is the 113kb size from the standard install? Maybe we could cut it down to only what we need? (via http://getbootstrap.com/customize/) If we go that route, we could also cut down on weight via gulp's minify.

nickmonad, line 8: Of the alternatives to bootstrap mentioned, purecss is my favorite

theMagicalKarp, line 8: I'm cool with bootstrap.  I actually stumbled across [uncss](https://github.com/giakki/uncss) the other day, which strips out all the unused css.  It plays well with [gulp](https://www.npmjs.org/package/gulp-uncss) too.  We can probably look more into this stuff when get around to optimization. 

`// digression`
Also speaking of optimizations, we can't just remove files from and image to make it smaller like I thought.  So deleting stuff after compilation is pointless.  It has to do with the way they are layered.
https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=docker%20image%20optimization

nickmonad, line 8: Ah interesting. Yeah we'll see what needs to happen as we reach that point. (in terms of both asset use and image size)

nickmonad:
@chfogerty @zach-taylor :eyes:

chfogerty:
+1

nickmonad:
Closing in favor of a client-to-repo refactor, see #8

PR #6, opened 2014-10-11
Backend tojson

cmbrose:
https://trello.com/c/mGit4Kp6/33-implement-back-end-portion-of-front-end-back-end-api-v0

The backend now returns JSON data instead of writing HTML.

Additionally, I added [Gorillamux](http://www.gorillatoolkit.org/pkg/mux) as the framework because it is very lightweight and easy to use.

The server currently only runs on localhost and will remain that way until we get better authentication.

In addition to the JSON move, this also serves as a basic API implementation with the following urls:

```
NOTE: prepend all urls with /api/v0.1
/
/images
/containers
/containers/create
/containers/create/<name>
/containers/stop/<id>
/containers/remove/<id>
```

To run:

```
go run backend.go certificates.go
```

The server will listen on :5000 and can be accessed through whatever method you prefer :)

There is also a test file associated with this addition to verify the backend's functionality.  To run:

```
go run backendTest.go
```

The output should hopefully be self explanatory.


cmbrose:
@ngmiller @chfogerty @zach-taylor @theMagicalKarp

backend/gotest/backend.go
```diff
@@ -0,0 +1,323 @@
...
+    r.HandleFunc(v + "/containers/stop/{id}", StopContainer).Methods("POST")
+    r.HandleFunc(v + "/containers/remove/{id}", RemoveContainer).Methods("DELETE")
+
+    http.Handle("/", r)
+    // Can use http.ListenAndServeTLS to enforce HTTPS once we have certificates
+    http.ListenAndServe(":5000", nil)
+}
+
+type DockerInfo struct {
+    Client      map[string]string
```
nickmonad, line 38: The map syntax is neat

backend/gotest/backend.go
```diff
@@ -0,0 +1,323 @@
...
+
+// Setting handles for the various Docker commands
+func main() {
+    v := "/api/v0.1"
+
+    r := mux.NewRouter()
+    r.HandleFunc(v + "/", GetInfo).Methods("GET")
+    r.HandleFunc(v + "/images", GetImages).Methods("GET")
+    r.HandleFunc(v + "/containers", GetContainers).Methods("GET")
+    r.HandleFunc(v + "/containers/create", CreateContainer).Methods("POST")
```
nickmonad, line 27: Using subrouters here may be helpful as the API support expands.

``` go
r := mux.NewRouter()
c := r.PathPrefix("/containers").SubRouter()

// -- /containers/create
c.HandleFunc("/create", CreateContainer).Methods("GET")

// -- /containers/create/{name}
c.HandleFunc("/create/{name}", CreateContainer).Methods("POST")
```

The documentation for the subrouter is quite lacking. It's buried in the overview on the site. My fundamental knowledge of go is lacking, so I'm not sure of a way to streamline the route to handler mapping and this construction. But, may be worth looking into

backend/gotest/backend.go
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    j, err := json.MarshalIndent(containers, "", "  ")
+
+    if err != nil {
+        fmt.Fprint(w, "{\"Error\":", err, "}")
+        return
+    }
+
+    fmt.Fprint(w, "{\n",
```
nickmonad, line 165: No need to return newlines for front-end consumption :)

backend/gotest/backend.go
```diff
@@ -0,0 +1,323 @@
...
+// Setting handles for the various Docker commands
+func main() {
+    v := "/api/v0.1"
+
+    r := mux.NewRouter()
+    r.HandleFunc(v + "/", GetInfo).Methods("GET")
+    r.HandleFunc(v + "/images", GetImages).Methods("GET")
+    r.HandleFunc(v + "/containers", GetContainers).Methods("GET")
+    r.HandleFunc(v + "/containers/create", CreateContainer).Methods("POST")
+    r.HandleFunc(v + "/containers/create/{name}", CreateContainer).Methods("POST")
```
nickmonad, line 28: I think `{name}` may need to be a query parameter like `.../create?name=thing`

backend/gotest/backend.go
```diff
@@ -0,0 +1,323 @@
...
+    w.Header().Set("Content-Type", "application/json")
+
+    client, err := GetDefaultClient()
+    if err != nil {
+        fmt.Fprint(w, "{\"Error\":", err, "}")
+        return
+    }
+
+    env, _ := client.Info() // json can't automatically marshal client
+    clientInfo := make(map[string]string)
```
theMagicalKarp, line 57: You can call `Map()` instead of iterating over the list. http://godoc.org/github.com/fsouza/go-dockerclient#Env.Map

backend/gotest/backend.go
```diff
@@ -0,0 +1,323 @@
...
+    for _, s := range *env {
+        kv := strings.Split(s, "=")
+        clientInfo[kv[0]] = kv[1]
+    }
+
+    images, _ := client.ListImages(false)
```
theMagicalKarp, line 63: I don't think aggregating images & containers from the `GetInfo` call is a good idea.  This is causing some overlap with `GetContainers` & `GetImages`.  Not to mention each docker call creates a new http request which comes with additional latency/overhead.

backend/gotest/backend.go
```diff
@@ -0,0 +1,323 @@
...
+    }
+    if len(apiImages) == 0 {
+        fmt.Fprint(w, "{}")
+        return
+    }
+
+    infos := make([]ImageInfo, len(apiImages))
+
+    for i, apiImage := range apiImages {
+        image, err := client.InspectImage(apiImage.ID)
```
theMagicalKarp, line 106: I think we should split up some responsibility here.  `GetImages` shouldn't inspect images automatically. It's also expensive to make this call n amount of times.  We should have a "inspect image" endpoint the client can hit.

backend/gotest/backend.go
```diff
@@ -0,0 +1,323 @@
...
+            continue
+        }
+
+        infos[i] = ImageInfo{Tags: apiImage.RepoTags, Image: *image}
+    }
+
+    j, err := json.MarshalIndent(infos, "", "  ")
+
+    if err != nil {
+        fmt.Fprint(w, "{\"Error\":", err, "}")
```
theMagicalKarp, line 118: We should be returning 500s on errors. Also we don't need a complete JSON object, a string should suffice. e.g.

``` Go
failureMessage, _ := json.Marshal(err.Error())
http.Error(w, failureMessage, 500)
```

backend/gotest/backend.go
theMagicalKarp, line 106: Same for `GetContainers`.

backend/gotest/backend.go
theMagicalKarp, line 165: Yeah and I think anonymous structures would be useful for what your trying to do.  e.g.

``` Go
response, _ := json.Marshal(struct {
    Containers []APIContainers
}{
    apiContainers,
})
```

theMagicalKarp:
The biggest thing I want to express is keeping a one to one relation with the docker api for basic container/image usage.  Essentially all we want is a simple proxy between a client and docker instances.

backend/gotest/backend.go
nickmonad, line 118: More specifically, we should be returning whatever the Docker remote API is returning for the targeted instance (Could be something in the 400 range as well.)

nickmonad, line 118: We'll have to be a little careful about how we differentiate between errors from our applications perspective, and errors from the target docker instance's perspective. It may be worthwhile to define a generic error response of the form,

```
{
    'error': 'docker',
    'message': 'fix it'
}
```

where the `error` value could potentially be `docker`, `host`, or `control` (or something to say "hey this is from us, not docker")

backend/gotest/backend.go
cmbrose, line 27: That's a good call Nick. I'll look into them.  It would definitely be nice to find a cleaner was of doing this than a long list of long lines.

backend/gotest/backend.go
cmbrose, line 165: That's a  really good idea @theMagicalKarp! I'll change things over to use that instead.  That'll take  away the burden of splicing together json bits too.

backend/gotest/backend.go
cmbrose, line 28: @ngmiller I had originally gone with this scheme following Docker's API, but I think that making the name an optional query parameter is a good direction to go for this (and it'll remove a route too!)

backend/gotest/backend.go
cmbrose, line 27: @ngmiller I think that I'll be changing this so that the URLs are of the form
`/containers/{name}/stop`
instead.  That may make the subrouters a little tougher depending on how they are implemented (whether or not I can still use things like {name} in a subrouter).

backend/gotest/backend.go
cmbrose, line 57: Good catch! Didn't see that

backend/gotest/backend.go
nickmonad, line 28: I had just thought wrong on how the Docker API was formatted in this case. I think sticking to their routes and conventions is important

backend/gotest/backend.go
cmbrose, line 106: @theMagicalKarp What happens with GetImages/Containers just depends on what kinds of data the frontend will be requesting at one time.  The client.ListImages data isn't particularly in-depth so I was thinking more along the lines of the basic management page wanting both the general Image and Container data at the same time.  Then when the user selects either Images or Containers they get much more in-depth info (which is the purpose of GetImages).  When they select an Image from there they get all the data we have (via an inspect endpoint I haven't made for this commit).

It just depends on when we want to get specific information for the user.  I can easily "shift" everything down a level and allow the frontend to decide exactly what info it wants and request it individually on that end rather than on my end.

nickmonad, line 106: It makes sense to keep everything as "atomic" as possible. Even if the front-end has to make 2 or 3 separate requests to load any initial data it needs, then that seems ok to me and keeps a decent level of separation

PR #7, opened 2014-10-11
GCE Discovery

theMagicalKarp:
### What
- set up a plugin module for routing request to specific plugins
- created a GCE module for handling GCE specific tasks

Currently lighthouse can detect other vms running inside a GCE project.  Specifically it detects vms that share the same project as itself. 
### Endpoints
- `/plugins/gce/vms` lists all the vms that lighthouse has discovered
- `/plugins/gce/vms/find` tells lighthouse to kick off a task to search for local vms, this requires an authorization code from google to properly run
- `/plugins/gce/authorize` redirects the user to googles gce two-factor login, which in turn once loged in, redirects the user back to our `/plugins/gce/vms/find` with an authorization code
### DiscoveredVM Struct
- **Name**  the name of the vm
- **Address** the internal private ip address of the vm, which is only exposed to other vms sharing the project
- **CanAccessDocker** this tells us whether or not we can access docker on the vm privately, we can detect this by hitting the `/v1/_ping/` endpoint on port 2375.
### Notes

Right now I'm storing data aggregated from the GCE plugin in a JSON file. I'll go through and change it later when we decide on a database.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,8 @@ ENV PATH $GOROOT/bin:$GOPATH/bin:$PATH
 
 WORKDIR /data
 
+RUN apt-get update
+RUN apt-get install -y mercurial
 
 # ----- Add private ssh key -----
 RUN mkdir -p /root/.ssh
@@ -60,6 +62,9 @@ WORKDIR /data/lighthouse
 # RUN git checkout -b some-branch origin/some-branch
 
 RUN go get github.com/fsouza/go-dockerclient
+RUN go get github.com/gorilla/mux
+RUN go get code.google.com/p/goauth2/oauth
+RUN go get code.google.com/p/google-api-go-client/compute/v1
 
 RUN npm install -g gulp
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,6 +63,9 @@ WORKDIR /data/lighthouse
 
 RUN go get github.com/fsouza/go-dockerclient
 RUN go get github.com/gorilla/mux
+RUN go get github.com/gorilla/sessions
+RUN go get github.com/gorilla/securecookie
+RUN go get github.com/bmizerany/pq
 RUN go get code.google.com/p/goauth2/oauth
 RUN go get code.google.com/p/google-api-go-client/compute/v1
 
--- a/backend/app.go
+++ b/backend/app.go
@@ -1,19 +1,44 @@
 package main
 
 import (
-    "./plugins"
-
+    "fmt"
     "net/http"
+    "time"
+
+    "./auth"
+    "./plugins"
 
     "github.com/gorilla/mux"
 )
 
 
+func LoggingMiddleware(h http.Handler) http.Handler {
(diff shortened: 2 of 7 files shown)


theMagicalKarp:
There is some overlap with pr https://github.com/ngmiller/lighthouse/pull/6.  I should just have to add this line, [#7 (diff)](https://github.com/ngmiller/lighthouse/blob/gce-discovery/backend/app.go#L15), to this file https://github.com/ngmiller/lighthouse/pull/6#discussion-diff-18739638R23. Once we have those changes merged.

nickmonad:
+1

backend/app.go
```diff
@@ -0,0 +1,19 @@
...
+    "github.com/gorilla/mux"
+)
+
+
+func main() {
+    r := mux.NewRouter()
+
+    plugins.Handle(r.PathPrefix("/plugins").Subrouter())
+
+    http.Handle("/", r)
```
nickmonad: @theMagicalKarp Could we extend the routing here to handle `/static/` for front-end assets? (https://trello.com/c/HQuz3cI9/42-remove-html-rendering-from-back-end-and-set-it-up-to-host-files-from-static)

nickmonad: Well, maybe we should just keep that as a separate PR. I don't think it will be too difficult :)

theMagicalKarp:
Jeeze Nick! What are ya doin! That's not in the scope of this ticket! :stuck_out_tongue:

zach-taylor:
+1

nickmonad:
+1

nickmonad:
@chfogerty @cmbrose Any more input here?

Dockerfile
```diff
@@ -63,6 +65,9 @@ RUN git clone git@github.com:lighthouse/lighthouse-client.git
 # Backend deps
 WORKDIR /data/lighthouse
 RUN go get github.com/fsouza/go-dockerclient
```
cmbrose, line 67: We're not using this anymore as far as I know

backend/app.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+    "./plugins"
+
+    "net/http"
+
+    "github.com/gorilla/mux"
+)
+
+
+func main() {
```
cmbrose, line 29: Is this going to fight the main in backend.go?

Do we need to merge them?

cmbrose:
Made a couple notes.  Otherwise +1!

backend/app.go
theMagicalKarp, line 29: When we get all the code merged I'll just put `plugins.Handle(r.PathPrefix("/plugins").Subrouter())` inside `bakcend.go` and delete this file.

Dockerfile
theMagicalKarp, line 67: Yup, this is just some code that got left in from a previous concept.  We can get rid of it in a "clean up" branch.

chfogerty:
+1

PR #8, opened 2014-10-12
Build

nickmonad:
https://trello.com/c/tY8JQ56q/44-docker-image-caching-breaks-development-branches

Here's an initial proof of concept as to what the build process could look like. It's currently untested but I just wanted to get things rolling on this.

From our discussion on trello, it seems like the fix for getting image builds to pull a fresh development branch is to have a separate script that clones the repository each time and grabs the specific branches. Since this operation won't be cached, it should reflect any recent changes.

That being said, I think we're looking at the container usage in the wrong way. I think this process slows down the iterative developments efforts I'd like to see when working on the front-end. The container build and run should really only be used when we're ready to actually deploy. From a development perspective, I should be able to just fire up the server locally (i.e. not in a docker instance), make some changes to the front-end, re-run gulp and see the changes instantly, without having to first (1) push changes to github (which requires a commit), and (2) restart some running container.

For a little more rigorous integration tests, it would be worthwhile going through the container build. And for that reason, we'll still need to pull in specific development branches.

I decided to put the front-end into a separate repo to facilitate testing. (with this, we could test two outstanding development branches simultaneously)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,14 +53,19 @@ RUN chmod 700 /root/.ssh/id_rsa
 RUN echo "Host github.com\n\tStrictHostKeyChecking no\n" >> /root/.ssh/config
 
 
-# ----- Setup enviroment -----
-RUN git clone git@github.com:ngmiller/lighthouse.git
-WORKDIR /data/lighthouse
-# You can play around with which branch you want the build use. Master is default.
-# RUN git checkout -b some-branch origin/some-branch
+# ----- Setup server enviroment -----
+# change this env variable to support a dev branch
+ENV back-branch master
+ENV front-branch master
 
+# Backend deps
+WORKDIR /data/lighthouse
 RUN go get github.com/fsouza/go-dockerclient
 
+# Frontend deps
 RUN npm install -g gulp
+
+# Command run on container run
+CMD ["/data/lighthouse/getcode.sh"]
 
 EXPOSE 5000
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,11 +52,13 @@ N2Pt8veiLc0cgXHeGE5rd+zXyeGcmra2W9pfptd49luerYEFQmIC\n\
 RUN chmod 700 /root/.ssh/id_rsa
 RUN echo "Host github.com\n\tStrictHostKeyChecking no\n" >> /root/.ssh/config
 
-
 # ----- Setup server enviroment -----
 # change this env variable to support a dev branch
 ENV back-branch master
 ENV front-branch master
+
+RUN git clone git@github.com:ngmiller/lighthouse.git
+RUN git clone git@github.com:lighthouse/lighthouse-client.git
 
 # Backend deps
 WORKDIR /data/lighthouse
@@ -65,7 +67,6 @@ RUN go get github.com/fsouza/go-dockerclient
 # Frontend deps
 RUN npm install -g gulp
 
+EXPOSE 5000
 # Command run on container run
 CMD ["/data/lighthouse/getcode.sh"]
-
-EXPOSE 5000
--- a/getcode.sh
+++ b/getcode.sh
@@ -14,6 +14,7 @@ git checkout back-branch
 cd /data/lighthouse-client
 git fetch origin
 git checkout front-branch
+npm install
 gulp build
 
 # Run server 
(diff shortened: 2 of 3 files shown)


nickmonad:
@theMagicalKarp @zach-taylor @chfogerty @cmbrose 
Sorry for the wall of text there, just wanted my intentions to be clear

nickmonad:
The sooner we get this sorted out, the better. Merging this into master and then master into the outstanding backend branches will only have a merge conflict in the Dockerfile, which should be straightforward to resolve

Dockerfile
```diff
@@ -53,14 +53,19 @@ RUN chmod 700 /root/.ssh/id_rsa
...
+WORKDIR /data/lighthouse
 RUN go get github.com/fsouza/go-dockerclient
 
+# Frontend deps
 RUN npm install -g gulp
 
-EXPOSE 5000
\ No newline at end of file
+# Command run on container run
+CMD ["/data/lighthouse/getcode.sh"]
```
nickmonad, line 72: There may be some permissions issues here. @theMagicalKarp @zach-taylor Is this the style of building you were thinking about?

getcode.sh
```diff
@@ -0,0 +1,21 @@
...
+# Pull backend
+cd /data/lighthouse
+git fetch origin
+git checkout back-branch
+
+# Pull frontend and build
+cd /data/lighthouse-client
+git fetch origin
+git checkout front-branch
+gulp build
```
nickmonad, line 16: This will copy the compiled (minified) assets up one directory into the backend/static folder.

Dockerfile
zach-taylor, line 72: This method of doing this will go and grab the code and build it every time you do docker run, which defeats one of the purposes of Docker. The lighthouse process should be ready to run when you do a docker run command, so the code should be grabbed during the docker build process.

In order to do this, we should do the following:
- git clone the repo
- build the Dockerfile (with the line "ADD . /data/lighthouse" to add the cloned source to the image)
- now you have the built image with compiled code, ready to run

Does that make sense? Wherever the docker image is being built will already need to have the whole repo to run the Dockerfile, so cloning will already have to be taken care of outside of the Dockerfile.

Dockerfile
```diff
@@ -53,14 +53,19 @@ RUN chmod 700 /root/.ssh/id_rsa
...
+
+EXPOSE 5000
```
zach-taylor: EXPOSE 5000 will need to be before the CMD line because CMD is the last thing that is run in a dockerfile.

Dockerfile
nickmonad, line 72: I'm not entirely sure we're on the same page. The issue with pulling any branch within the Dockerfile (and thus, within the docker image build) is that layer of the filesystem gets cached. So, if I need to make a change and test it, the docker image build will not pull the changes I just pushed to github because as far as its concerned, nothing changed

nickmonad, line 72: For production deploys, one would fix this by pulling specific tags. Then, docker will see that layer of cache is invalidated when the Dockerfile tag number changes, and it will forced a repull. But, tagging for development branches doesn't make sense

nickmonad, line 72: This still gives us the benefit of docker images; we just need to put our own volatile changes on top

zach-taylor, line 72: Well if we're looking to grab the latest code every time we run the container, we could at least place the git clones (with the branch set as an env variable), and then just do a git pull in the getcode.sh. That way we aren't grabbing the entire repository every time the container is started/restarted, just the changes.

nickmonad, line 72: Oh yeah good call! Updating now

nickmonad:
Another potential fix: https://docs.docker.com/userguide/dockervolumes/#mount-a-host-directory-as-a-data-volume

theMagicalKarp:
@ngmiller Yeah I agree, the Dockerfile only seems useful for deployment.  For working locally I would recommend just running sans Docker for now.  That way aren't being slowed down by pushing up to github.
+1

nickmonad:
It's definitely not impossible to use Docker as a development process (see: https://github.com/wsargent/docker-cheat-sheet#tldr-i-just-want-a-dev-environment), but for what we have now, I don't think it's strictly necessary, as long as the backend can run locally :) For development, we could probably just point the backend to a local boot2docker vm as the target host instance to control

nickmonad:
Code reviewed at group meeting on 10/16

PR #9, opened 2014-10-15
Setting up User Models & Authentication

theMagicalKarp:
### What

Setup a way to connect to a Postgres database and store user information in our app as well as endpoints to authenticate users.  Also created a config directory that can be modified inside a container or read via shared volume.  This should allow other people to configure lighthouse for more flexibility.  Right now `config/auth.json` is the file that determines admin credentials and `config/plugins/gce.json` the gce oath configuration.
### Postgres

Right now Postgres is setup to be either ran from via Postgres container or locally.  On start up it looks for the container first.  If the container DNE then it looks for a locally running Postgres server.  Hopefully this should help with local development.
### Endpoints
- `/login` A POST endpoint that accepts a json object with two attributes `Email` & `Password`. The endpoint returns a bool that can be used for success feedback.
- `/logout` A GET endpoint that strips authentication from a user's session cookie.
### Running in Docker

Here are a few useful arguments to use when running the app inside Docker.
##### Linking to Postgres Container

```
--link name-of-postgres-container:postgres
```
##### Setting working directory to app.go

```
-w /data/lighthouse/backend
```
##### Sharing config volume

```
-v /home/rob.sheehy/config:/config
```
### Hacking Challenge

So I thought it would be fun to create a little hacking challenge for our team.
Currently I'm running a modified version of this image on good ole dreamy-hoover @ http://146.148.34.150/ with a single user `test@gmail.com`.  I've placed a "secret word" that can be accessed from http://146.148.34.150/secret, however it sits behind the authentication I built in this branch.  The goal is to get that "secret word" any way you can .  I won't tell you the password and I'm also trusting you won't ssh into the vm  and read lighthouse config to cheat ;). 
Also here is a simple script to check passwords for `test@gmail.com`'s account via browser console.

``` JavaScript
var xhr = new XMLHttpRequest();
xhr.open("POST", "/login");
xhr.setRequestHeader('Content-Type', 'application/json');
xhr.send(JSON.stringify({'Email': 'test@gmail.com','Password': 'xxxxxxxxxx'}));
```

If you guys can figure out the "secret word" I probably need to do some rethinking, otherwise I'll probably sleep a lot better night knowing we are reasonably secure.
### Side Notes
- I don't like the way `AuthMiddleware` is setup, that is the URL ignore argument part of it. Wrapping the main mux was the only way I could figure out how to write middle ware for Gorilla.  I think this is where other other web frameworks shine such as [martini](https://github.com/go-martini/martini#middleware-handlers) or [goji](https://github.com/zenazn/goji/blob/master/example/middleware.go) shine over Gorilla.
- I created `LoggingMiddleware` so that we can view incoming request on our server.
### What's next
- buy a domain
- get a signed certificate
- ???
- profit

### DIFF
--- a/backend/app.go
+++ b/backend/app.go
@@ -1,19 +1,44 @@
 package main
 
 import (
-    "./plugins"
-
+    "fmt"
     "net/http"
+    "time"
+
+    "./auth"
+    "./plugins"
 
     "github.com/gorilla/mux"
 )
 
 
+func LoggingMiddleware(h http.Handler) http.Handler {
+    return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+        start := time.Now()
+
+        h.ServeHTTP(w, r)
+
+        end := time.Now()
+        latency := end.Sub(start)
+        fmt.Printf("%s %12s %s %s\n",
+            end.Format("2006/01/02 - 15:04:05"), latency, r.Method, r.URL)
+    })
+}
+
+
 func main() {
+    fmt.Println("Starting...")
     r := mux.NewRouter()
+
+    auth.Handle(r)
 
     plugins.Handle(r.PathPrefix("/plugins").Subrouter())
 
-    http.Handle("/", r)
+    ignoreURLs := []string{"/", "/login", "/logout", "/plugins/gce/vms/find"}
+    app := auth.AuthMiddleware(r, ignoreURLs)
+
+    http.Handle("/", LoggingMiddleware(app))
+
+    fmt.Println("Ready...")
     http.ListenAndServe(":5000", nil)
 }
--- a/config/auth.json
+++ b/config/auth.json
@@ -1,0 +1,6 @@
+{
+    "Admins": [
+        {"Email": "admin@gmail.com", "Password": "admin"}
+    ],
+    "SecretKey": "I'm a secret key..."
+}
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,6 +63,9 @@ WORKDIR /data/lighthouse
 
 RUN go get github.com/fsouza/go-dockerclient
 RUN go get github.com/gorilla/mux
+RUN go get github.com/gorilla/sessions
+RUN go get github.com/gorilla/securecookie
+RUN go get github.com/bmizerany/pq
 RUN go get code.google.com/p/goauth2/oauth
 RUN go get code.google.com/p/google-api-go-client/compute/v1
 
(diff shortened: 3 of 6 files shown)


theMagicalKarp:

nickmonad:
Wow, that was fast. Also, http://146.148.34.150/ is currently 404'ing
EDIT: nevermind :)

nickmonad:
A diagram of the auth architecture would be nice for our purposes as well as the upcoming design document

zach-taylor:
Wow @theMagicalKarp you are super speedy!
Regarding the cert and domain. If we use CloudFlare, we can manage the domain and get free SSL, plus some nice security and CDN functionality :)

I will look into pwning the auth.

theMagicalKarp:
@ngmiller 
- Yeah nothing is listening to the root of our app :(, however all the other endpoints should be working.
- Sure I'll try to document and diagram something before we meet tomorrow.

@zach-taylor 
- +1 CloudFlare
- Bring it, also if you see any bad practices let me know.

theMagicalKarp:
Also if you guys somehow get authenticated you can checkout the gce plugin in action http://146.148.34.150/plugins/gce/vms

config/auth.json
```diff
@@ -0,0 +1,6 @@
+{
+    "Admins": [
+        {"Email": "admin@gmail.com", "Password": "admin"}
```
nickmonad, line 3: Just looking over this again (albeit quickly) - Do we have a plan for creating administration accounts through the system in place here? Probably not a good idea to hardcode this in. At very least, we need a password different from "admin" :)

backend/app.go
```diff
@@ -1,19 +1,44 @@
...
+func LoggingMiddleware(h http.Handler) http.Handler {
+    return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+        start := time.Now()
+
+        h.ServeHTTP(w, r)
+
+        end := time.Now()
+        latency := end.Sub(start)
+        fmt.Printf("%s %12s %s %s\n",
+            end.Format("2006/01/02 - 15:04:05"), latency, r.Method, r.URL)
```
nickmonad, line 24: 2006 ? EDIT: oh is that just a template of some sort?

theMagicalKarp, line 24: Yeah it kinda bothers me too but it's the way time formatting is done in go. http://golang.org/pkg/time/#Time.Format

config/auth.json
theMagicalKarp, line 3: Yeah, the idea is that the person running the image should have their own `auth.json` file with their own configuration inside a shared volume.  So theoretically this configuration should never be ran, unless you're running on localhost.  Does that answer your question? :/

nickmonad, line 3: Oh yeah I see. Makes sense!

chfogerty:
+1.

Raises a few questions/things to do, but they're outside the scope of this PR. I'll bring them up next meeting.

nickmonad:
+1

PR #10, opened 2014-10-18
Server update and custom handlers

cmbrose:
https://trello.com/c/mGit4Kp6/33-implement-back-end-portion-of-front-end-back-end-api-v0

This is an initial draft of my vision of handling the Docker API as discussed in our last team meeting.

Requests can be sent to a Docker instance through the endpoint
`/api/v0.1/__<Host>__/d/<DockerEndpoint>`

This currently should fully support GET, POST, PUT, and DELETE requests.

Testing:
- Start a Docker instance which listens on an open port: `sudo docker -H localhost:5001 -d &`
- Start the backed: `go run backend.go`
- Hit port 5000 on localhost with the request as formatted above

A basic custom handler interface has also been created. This is very subject to change as we hash out requirements there.


cmbrose:
@ngmiller @chfogerty @theMagicalKarp @zach-taylor

backend/server/backend.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+/*
+    Sets endpoints for the server and begins listening for requests
+*/
+func main() {
+    baseRouter    := mux.NewRouter()
+    versionRouter := baseRouter.PathPrefix("/api/v0.1").Subrouter()
+
+    // TODO - Host may contain /'s, need a cleaner way to support this
```
cmbrose: @ngmiller @theMagicalKarp @zach-taylor @chfogerty 
Any thoughts on this?

backend/server/handlers/handlers.go
```diff
@@ -0,0 +1,71 @@
...
+      - r.ParseForm() will have been called prior to calling the handler
+
+      - it will not be asked to rollback if it has not been called
+        already for a particular request
+
+      - it will not be asked to rollback if it caused an unrecoverable
+        error for a particular request (i.e. initiated a rollback)
+
+      - its error message will be returned to the client during failure
+*/
```
cmbrose, line 43: @chfogerty 
This probably affects you the most.  Thoughts?

backend/server/handlers/dockerRequestHandler.go
```diff
@@ -0,0 +1,56 @@
...
+    in the request's 'Payload' field is also forwarded.
+
+    RETURN: true on success, false otherwise
+*/
+func DockerRequestHandler(w http.ResponseWriter, r *http.Request) *HandlerError {
+    vars := mux.Vars(r)
+    host := vars["Host"]
+    url := "http://" + host + "/" + vars["DockerURL"]
+
+    payload := r.Form["Payload"]
```
nickmonad: #nitpick JSON keys are typically camelCased with the first letter lowercase

nickmonad: but this could just be `payload`

backend/server/backend.go
nickmonad: Generally speaking, REST urls have an ending `/` to denote collections of objects, while there will be no trailing slash for a specific object to manipulate. (I think that's what you mean :) )

backend/server/backend.go
```diff
@@ -0,0 +1,72 @@
...
+    hostRouter    := versionRouter.PathPrefix("/__{Host:.*}__")
```
nickmonad: Any particular reason for the double underscores?

backend/server/handlers/handlers.go
```diff
@@ -0,0 +1,71 @@
...
+
+type HandlerError struct {
+    StatusCode int
+    Cause string
+    Message string
+}
```
nickmonad, line 54: Will this get rendered as (?)

```
{
    'StatusCode': ...,
    'Cause': ...,
    'Message': ...,
}
```

If so, it's probably best to stick with camelCase (first letter lower)

backend/server/handlers/dockerRequestHandler.go
nickmonad: Off the top of my head, I think there is a distinction between an HTTP form post and a body. In which case, we'll be posting JSON objects into the body

backend/server/handlers/dockerRequestHandler.go
```diff
@@ -0,0 +1,56 @@
...
+    var payloadBody []byte
+
+    if len(payload) > 0 {
```
nickmonad: Will this throw an exception if `r.Form["Payload"]` doesn't exist?

nickmonad:
If we can, it would be nice to have some kind of global request wrapper that will allow us to intercept any errors that crop up during the request/handler chain. That way, we could format the error response how we want and not accidentally send back debugging information and raw server responses to the client.

Otherwise, looks like a great start!

backend/server/backend.go
cmbrose: I mentioned it in the comment right above the line.  If the host url can contain /'s we need a way to detect exactly when the host field begins and ends.

Double underscores isn't really a good way of supporting this, but it works until we can come up with a clean method.

backend/server/handlers/dockerRequestHandler.go
cmbrose: If `'Payload'` doesn't exist, payload should be set to "".  (I was able to perform both GET and POST requests that didn't have `'Payload'` and didn't have any issues.

cmbrose:
@ngmiller Are you thinking almost along the lines of a try-catch?  Otherwise if you mean sending back failure messages in the handlers, that is handled already.  If a handler has an error it should return a handler.HandlerError which will then be written to the response.

backend/server/handlers/handlers.go
cmbrose, line 54: @ngmiller See the `WriteError` function in backend.go for specifics of how these are used.

Basically the return will look like:

```
HTTP/1.1 <StatusCode>

{
    'Error': <Cause>,
    'Message': <Message>
}
```

backend/server/backend.go
zach-taylor: I don't believe hostnames can contain /'s

nickmonad: Oh I see. Sorry, I read host as "URL" for some reason. We could just allow the user to set some alias for the host they are targeting. So, maybe their production gce instance would be labelled 'prod-gce' and we map that to the actual URL or IP address. I don't foresee the host target having forward slashes in the URL, but an alias system would handle all cases.

backend/server/handlers/handlers.go
cmbrose, line 54: Which, we can always turn these into the `'Error'` and `'Message'` into camelCase, but the fields in HandlerError should remain PascalCase.  (Look into [exported types](https://golang.org/ref/spec#Exported_identifiers))

nickmonad:
@cmbrose Yeah something like a global try-catch. From what I can tell, Go doesn't support exceptions in the traditional style, but some kind of middleware between the incoming request and the handlers may prove to be beneficial. Things like that are mainly implemented for security reasons, to be sure that a predefined response not leaking any system information is sent back in the worst case. But, if every handler is set up that way, then we should be good

backend/server/handlers/dockerRequestHandler.go
cmbrose: Can someone verify this?  It's an easy fix, just don't want to change it and then change it back.

backend/server/handlers/handlers.go
nickmonad, line 54: I guess it's just a matter of taste. I personally prefer camelCase for JSON, but it's not a big deal now.

backend/server/handlers/dockerRequestHandler.go
nickmonad: http://golang.org/pkg/net/http/#Request - I think it should be body. Looks like the form does some footwork to include the query parameters as well.

backend/server/backend.go
nickmonad: ^ ignore this

backend/server/handlers/dockerRequestHandler.go
```diff
@@ -0,0 +1,56 @@
...
+    } else {
+        payloadBody = nil
+    }
+
+    req, err := http.NewRequest(r.Method, url, bytes.NewBuffer(payloadBody))
+    if err != nil {
+        return &HandlerError{500, "control", "Failed to create " + r.Method + " request"}
+    }
+
+    // TODO - better client
```
nickmonad, line 32: Trello card? :) I've made a "backlog" column for things we want to consider in future sprints but not necessary for MVP

backend/server/handlers/handlers.go
cmbrose, line 54: @ngmiller Sorry, I don't mean to seem like I'm disregarding your tastes.

HandlerErrors should never see the light of day in terms of moving onto the client in JSON and if I make them camelCase they will become private and basically unusable.  That is why I can't change their casing.

The only reason I have JSON being PascalCase is because these Error and Message fields are automatically PascalCase and I followed that convention for the rest of the JSON.

I have zero qualms about converting all input and output JSON to camelCase.

cmbrose:
@ngmiller Take a look at the Custom Handler contract in handlers.go.

Handlers are expected to catch all their own errors and handle them according to a fairly strict convention that should protect us from sending junk data to the client.

But I guess **expected** is not **will**.

backend/server/handlers/handlers.go
nickmonad, line 54: > Sorry, I don't mean to seem like I'm disregarding your tastes.

Haha don't worry. I think I came across as too prescriptive. I knew your reasons were technical/objective. I really don't mind either way. I'm suprised go's `json.marshal` doesn't move keys to camelCase. If it's easy to implement, it'd be nice, but ultimately a non-issue

backend/server/backend.go
cmbrose: > I don't believe hostnames can contain /'s

I'm probably misunderstanding our intentions here then.

How to we get the Docker instance to forward requests to?  I had essentially been concatenating the `{Host}` and `{DockerURL}` values for the forwarded HTTP request address.

nickmonad:
@cmbrose Re: custom handlers - Ok that's fair. As long as we stick to the interface/convention, we should be set

backend/server/handlers/dockerRequestHandler.go
cmbrose, line 32: Added :)

backend/server/backend.go
nickmonad: If we don't go with a host alias map (i.e. `gce.prod` => `some.gce.domain.com`) then I think it's fair to not allow `http://` in the host portion of the API url. So we'll have to add it ourselves. (Unless the underlying request library adds it automatically) - Is that what you mean?

nickmonad:
@cmbrose I'm on gitter right now if you wanna chat

Dockerfile
```diff
@@ -62,11 +62,11 @@ RUN git clone git@github.com:lighthouse/lighthouse-client.git
...
-RUN go get github.com/fsouza/go-dockerclient
+RUN go get github.com/gorilla/mux
 
 # Frontend deps
 RUN npm install -g gulp
 
 EXPOSE 5000
 # Command run on container run
-CMD ["/data/lighthouse/getcode.sh"]
+CMD ["/data/lighthouse/getcode.sh"]
```
cmbrose: @theMagicalKarp can you look at this and make sure it will still work for you?

cmbrose:
@ngmiller @theMagicalKarp @zach-taylor @chfogerty this should be ready to go for local testing.

Dockerfile
theMagicalKarp: Looks good to me.

theMagicalKarp:
@cmbrose Overall this looks like a good start.
I just had some nitpick stuff that's probably more personal preference than anything else.
I'm not a fan of the project structuring.  Tell me if I'm wrong, but I think your leaning towards something like this.

```
/backend
  /server
    backend.go
    /models
      userModel.go
      pluginModels.go
    /handlers
      dockerHandler.go
      handlers.go
      authHandler.go
      gcePluginHandler.go
    /hosts
      lookup.go
```

Not that I completely disagree with that structure, I tend to lean towards a more modular setup.

```
backend.go
/auth
  handler.go
  models.go
/plugins
  handler.go
  /gce
    handler.go
    models.go
/docker
  handler.go
  models.go
```

Such that `handler.go` would define front-end interaction and `models.go` would handle all the database calls.
IMO I think this helps project scope, consistency, scalability, and readability.
For instance if I wanted to work on stuff pertaining to the gce plugin I would just have to cd to `/plugins/gce` and checkout the database models in `models.go` and see how the front-end is interacting with those models inside `handler.go`.  Everything I would want to know about gce plugins should be contained in the `/plugins/gce` dir.

Like I said this is probably personal preference over anything.  As a group we should establish some project patterns/standards soon.

Otherwise everything looks good +1, lets start merging some branches!

cmbrose:
@theMagicalKarp that definitely sounds like a more logical system to me!  We can hash it out in our meeting this week, but probably do that in a different PR after the outstanding ones go away.

nickmonad:
+1

chfogerty:
+1

PR #11, opened 2014-10-25
Setting up Docker the way God intended!

theMagicalKarp:
### What

So yesterday I had a sudden "Dockerfile epiphany" which opened my mind to how wrong we are using our Dockerfile.  This all started when I was reading the recent [article](http://jonathan.bergknoff.com/journal/building-good-docker-images) that Nick posted which peaked my curiosity in how we can improve our Docker build.  Then I stumbled across the [google/golang](https://registry.hub.docker.com/u/google/golang/) image, and things just started making sense.  As it turns out we can copy local code directly from our dev machines into a docker container, with the `ADD` command.  This meant several things.
- Changes no longer rely on git branching.
- We don't need node, since static resources can be built outside a image and be pushed into the image.
- We can properly setup our go packaging without messing with our local gopath.
- We won't need to store a private key in our Dockerfile. :sweat_smile: 
##### Build Local Image

`docker build -t lighthouse .`
##### Boot Postgres Container

`docker run --name some-postgres -d postgres`
##### Start Lighthouse

`docker run -t -i -p 5000:5000 --link some-postgres:postgres lighthouse`
### Notes

So in the Dockerfile I run go get on all our dependencies before I pull the local code.  This isn't necessary because `go get github.com/ngmiller/lighthouse` will actually do this part for us.  What I've found though is if we do all the go gets before we add our code all the libraries are cached.  This way every time we do a build we don't have to re-"go-get" the libraries in turn making builds much faster.

I also set up our packaging in the way go packages are meant to be setup, that is I used `github.com/ngmiller/lighthouse`.

Plus this new image is nearly half the size of our current one.

rob ford


theMagicalKarp:
@ngmiller @chfogerty @cmbrose @zach-taylor If this gets merged that means we can go back to focusing on developing locally with Docker. :dancer:

Dockerfile
```diff
@@ -1,78 +1,17 @@
...
+FROM google/golang
 
 MAINTAINER lighthouse
 
-# ----- Install go -----
-WORKDIR /root
-RUN \
-  mkdir -p /goroot && \
-  curl https://storage.googleapis.com/golang/go1.3.1.linux-amd64.tar.gz | tar xvzf - -C /goroot --strip-components=1
+WORKDIR /gopath/src/github.com/ngmiller/lighthouse
```
nickmonad, line 5: How does this path get set up? Will docker automatically create this tree if the directories don't exist?

nickmonad:
Looks awesome! Could we add a script that does the front-end building? It seems like that sequence still needs to be done manually

cmbrose:
After just spending hours wrestling Postgres locally...



+1

Dockerfile
theMagicalKarp, line 5: Seems that way. ¯ \_(ツ)_/ ¯

theMagicalKarp:
@ngmiller Do you mean where we clone lighthouse-client and run the gulp build?
 I almost want to say that we should...
- combine the 2 repos
- add code to static serve the client in `lighthouse.go`
- run gulp build locally
- build the docker image

I don't know though, I haven't taken a look at lighthouse-client lately.  We could possibly look into setting it up as subproject too.  Otherwise I'm fine with setting up that client building script. What do you guys think?

nickmonad:
@theMagicalKarp Combining the front-end back into the main repo wouldn't be too bad, since the directory structure is starting to solidify. But, keeping them separate facilitates testing two development branches easily. Anyway, it seems like the Dockerfile is written under the assumption that the repo is already cloned locally. If that's the case, then lighthouse-client should be as well, and it should just be a matter of running `gulp prod build` from lighthouse-client root. The current build process for the client makes an assumption as to the location of the static assets in the back-end, but that's easily changed

zach-taylor:
@theMagicalKarp I like what you've done here. looks really good. Does Go have a dependencies file convention (e.g. Gemfile in Ruby, requirements.txt in Python)

@ngmiller @theMagicalKarp we could git submodule the front-end code if that would work. I think we should keep the repos separate for front-end and back-end mainly because, like Nick said, the development will be separate.

zach-taylor:
@theMagicalKarp we could investigate adding a fig.yml file to the repo. http://www.fig.sh/ looks pretty slick. It would make local development a lot simpler. I think the only command you have to run is "fig up" when you want to run everything (app container + postgres container + whatever). It handles the linking of containers as well.

theMagicalKarp:
@zach-taylor Go doesn't really need to have package manager since `go get github.com/ngmiller/lighthouse` will detect external dependencies automatically and install them.  The only reason I'm manually running go get on our external dependencies is so that we can cache them in the image, before add all the lighthouse code.  This is because that `ADD` line in the Dokerfile is going to break the image cache a lot, meaning that anything after our `ADD` is adding time to the average build time.

Alright, in the case we are keeping them separated, I'll research into setting it up as a subproject.  <sup>I might not know what I'm doing ;)</sup>

Yeah I've been curious about using something like fig ever since we brought Postgres into the equation.  Besides [fig](https://github.com/docker/fig/) there's also [tug](https://github.com/nitrous-io/tug) and [crane](https://github.com/michaelsauter/crane).  I'm probably leaning towards fig though.

zach-taylor:
@theMagicalKarp sounds good.
@cmbrose did you mean to close this PR?

cmbrose:
@zach-taylor oops... must have hit the wrong button... my bad.

zach-taylor:
@cmbrose All good. I did the same thing lol

nickmonad:
+1

chfogerty:
+1

PR #15, opened 2014-11-03
Refactor Plugins to Provider module

theMagicalKarp:
### What

Refactored the way we communicate with available vms on a private network.  We now guess where lighthouse is running and interface with the provider in a particular way.
### Endpoints
- `/provider/vms` A GET endpoint that returns a json array of private vms available. [example here](http://146.148.78.229/provider/vms)
- `/provider/which` A GET endpoint that indicates which provider the backend is configured to. [example here](http://146.148.78.229/provider/which)
### Provider "Flavors"
- **gce:** This is applicable if we can detect google metadata endpoints.  We can then assume that we have access to the gce api and make calls to it to discover vms.
- **local:** This is only applicable if the `DOCKER_HOST` environment variable exists.  We can then assume that we have access to the local Docker daemon through the `DOCKER_HOST` address.
- **unknown:** This is only applicable if no other provider flavor can be determined.  The intentions of the flavor is to only return empty responses.
#### VM
- **Name:** The human readable name of the vm instance
- **Address:** The private address of the vm.
- **CanAccessDocker:** Indicates if lighthouse can make api calls to the hosted Docker daemon.
### Notes
- Instead of using Google Two-Factor Authentication, we now require that vm running lighthouse has elevated privileges to view and modify gce instances.  It's a setting that's set when creating a gce instance. This should make it much easier for lighthouse to interact with the compute engine api.
  screen shot 2014-10-21 at 8 52 34 am


robertsheehy-wf:

provider/models/models.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+type Provider struct {
+    Name string
+    IsApplicable func() bool
+    GetVMs func() []*VM
+}
+
+func PingDocker(address string) bool {
+    pingAddress := fmt.Sprintf("http://%s:2375/v1/_ping", address)
```
nickmonad: I think the VM struct should have a port associated with it that's set during configuration in the event Docker is not listening on 2375. And maybe an API version as well.

lighthouse.go
```diff
@@ -67,7 +67,7 @@ func DockerHandler(w http.ResponseWriter, r *http.Request) {
 
 
 func main() {
-    fmt.Println("Starting...!!!")
+    fmt.Println("Starting...")
```
nickmonad: Does go have any built-in logging capability? It'd be nice to have timestamped messages in a configurable log file.

nickmonad:
Looks great so far. +1 with updated [API docs](https://github.com/lighthouse/lighthouse/wiki/API-Design). The other comments are smaller concerns now, but just stuff to consider

nickmonad:
Also I'm claiming patent rights on "de-millerize"

lighthouse.go
robertsheehy-wf: We can use [this](http://golang.org/pkg/log/) to do the something like the following...

``` Go
package main

import (
    "os"
    "log"
)

func main() {
    logging := log.New(os.Stdout, "", log.Ldate | log.Ltime | log.Lshortfile)
    logging.Println("starting")
}
```

which outputs...

```
2014/11/03 16:08:05 rob.go:10: starting
```

provider/models/models.go
robertsheehy-wf: That seems fair, when I get time I'll add a `port` field to the VM struct. For now though I'll probably just default them to `2375`.  Same for `version` which will probably be defaulted to `v1`.

lighthouse.go
nickmonad: :+1:

provider/models/models.go
nickmonad: Sounds good!

robertsheehy-wf:
@ngmiller Created a logging module and updated the VM model.

nickmonad:
:+1:

theMagicalKarp:
@chfogerty @zach-taylor @cmbrose Speak now or forever hold your peace!

cmbrose:
Looks good! +1

provider/providers/gce/serviceaccount.go
```diff
@@ -0,0 +1,173 @@
+// Copyright 2013 The goauth2 Authors. All rights reserved.
```
chfogerty: I believe we'll need to include their license. Is that in here already?

theMagicalKarp: @chfogerty Idk, if that's the case, do we need to include all the libraries we use? I would do a `go get` on this, but because we needed `metadataServer = "metadata.google.internal"` and not `metadataServer     = "metadata"`.  This won't work inside a Docker container otherwise.

theMagicalKarp:
@chfogerty Fixed

PR #17, opened 2014-11-11
Backend unit tests

cmbrose:
The new files here are unit tests for the backend handlers and this is roughly how I envision the unit tests for the backend to look.

To run the tests, just navigate into a directory with a test file in it and run `go test`


cmbrose:
@chfogerty @theMagicalKarp what do you think of this style for our future tests?

@ngmiller @zach-taylor do you guy shave any comments?

handlers/dockerRequestHandler_test.go
```diff
@@ -0,0 +1,183 @@
...
+    if f != nil {
+        useFunc = *f
+    } else {
+        useFunc = func(http.ResponseWriter, *http.Request) {}
+    }
+
+    server := httptest.NewUnstartedServer(http.HandlerFunc(useFunc))
+    server.Config.Addr = "/"
+
+    go func() {
```
theMagicalKarp: Why is this async? I have no idea what's going on, but it seems like there could be some race conditions.

theMagicalKarp:
Pro tip `go test github.com/lighthouse/lighthouse/...` or `go test ./...` tests all the lighthouse packages in one shot. ^_^
Also for those testing this out, I had to manually install the testing framework since I think `go get github.com/lighthouse/lighthouse` ignores dependencies for test files.  So all you should need to do is `go get github.com/stretchr/testify`.
Otherwise most of this stuff looks fine.

cmbrose:
@theMagicalKarp the async was an artifact of an old testing method I had tried and decided against.  It's gone now.  You brought up a good point about the last of comments in these tests though.  I'll get some added soon!

handlers/handlers.go
```diff
@@ -159,8 +155,8 @@ func GetHandlerInfo(r *http.Request) HandlerInfo {
     vars := mux.Vars(r)
     var info HandlerInfo
 
-    info.Host = hosts.AliasLookup(vars["host"])
-    info.DockerEndpoint = vars["dockerURL"]
+    info.Host = hosts.AliasLookup(vars["Host"])
+    info.DockerEndpoint = vars["DockerURL"]
```
cmbrose, line 159: @ngmiller @zach-taylor probably the biggest thing here for you guys is that all the JSON from the handlers will now be PascalCase (and the backend will expect the same of incoming data as well).  This is because Docker's API uses PascalCase.

nickmonad, line 159: Sounds good

nickmonad:
+1

theMagicalKarp:
+1

cmbrose:
@chfogerty last chance to stop a merge on this

chfogerty:
+1

PR #21, opened 2014-11-21
Provider adding

cmbrose:
So this one is a bit of a doozy...

**Primary additions**
1. Hostname aliasing
2. Permissions for Docker requests

**API additions**
`GET /api/v0.1/users/permissions/vms`
- lists the IP address of all Docker instances the current user can access (i.e. those that won't be blocked)

**Why this is _so_ massive**
1. Added a postgres package to ease database work:
   - storage here is in key-value pairs where key is a string and value is anything (it's turned into JSON) 
   - all keys are unique (such as emails) 
   - database_testing_utils.go helps mock database calls to lighten dependencies in unit tests
2. Added session package in order to:
   - make auth fields (like email) available everywhere without worrying about circular dependencies
   - allow us to store non-auth values
3. User class moved to its own package so that everything can access it

**Testing**
Testing for non-Docker endpoints can be done as usual.

With aliasing, we have to do a little work to make Docker listen to us:
`docker -H 0.0.0.0:5001 -H unix:///var/run/docker.sock -d &`
This makes the docker daemon listen on port 5001 as well as its usual socket.

Then find Docker's IP using `ifconfig`.  It should default to `172.17.42.1`, but if not you will need to change this value in `config/permissions.json`.

Now you're ready to make requests!!

The alias of the local Docker instance is `boot2docker`, so the request format is:
 `localhost:5000/api/v0.1/boot2docker/d/{DOCKER_REQUEST}`


nickmonad:
Nice work. Seems like a lot of good stuff in here. I tried my hand at a little SQL injection and we're good on the host front due to the nature of the permissions check:

screen shot 2014-11-21 at 3 03 57 am

... but we'll have to sanitize the input somehow on API calls that accept user input, hit the database, and don't do a check similar to host permissions (perhaps on AddAlias in aliases.go)

nickmonad:
http://xkcd.com/327/ :)

cmbrose:
@theMagicalKarp @chfogerty @zach-taylor

cmbrose:
@ngmiller it's probably blind luck that the injection didn't work ;) I haven't done anything to try and stop it yet.  If you want to really try and break it, the queries are in `postgres/postgres.go` and it's probably best to sanitize there instead of in everything that uses postgres.

robertsheehy-wf:
Yeah I 100% agree with Nick. The pattern of including the args in the `Exec` calls sanitizes the input for us.

``` Go
db.Exec("INSERT INTO users (email, salt, password) VALUES (($1), ($2), ($3))",
        email, salt, saltedPassword)
```

Otherwise upon first glance this stuff looks good.

robertsheehy-wf:
https://github.com/lighthouse/lighthouse/commit/432e8e289dc2e231ea157034f5257b9027b985e5

cmbrose:
@theMagicalKarp I had that style originally, but I couldn't get it to work...

``` go
this.db.Exec(`CREATE TABLE ($1) (($2) text UNIQUE PRIMARY KEY, ($3) json);`,
        this.table, keyColumn, valueColumn)
```

where `this.table` is "users" gives a runtime error:

```
pq: R:"parserOpenTable" S:"ERROR" C:"42P01" M:"relation \"users\" does not exist" P:"13" F:"parse_relation.c" L:"984"
```

cmbrose:
@theMagicalKarp false alarm, I've solved the issue.  The solution was to only use the `($1)` for values only and use `%s` for table name and column:

``` go
query := fmt.Sprintf(`SELECT %s FROM %s WHERE %s = ($1);`,
    valueColumn, this.table, keyColumn)

row := this.db.QueryRow(query, key)
```

robertsheehy-wf:
@cmbrose Good catch. Yeah generally I think it's safer to statically declare tables and fields over dynamically creating them in SQL.  Otherwise in this case I think we should be fine.

I'm also surprised this works without the [postgres sql driver](https://github.com/lib/pq). 
Some tools inside pq would also give us the ability to sanitize the table and field names. http://godoc.org/github.com/lib/pq#QuoteIdentifier
For example...

``` Go
query := fmt.Sprintf(`SELECT %s FROM %s WHERE %s = ($1);`,
    pq.QuoteIdentifier(valueColumn), pq.QuoteIdentifier(this.table), keyColumn)
```

I'll leave that at your discretion though.

cmbrose:
@theMagicalKarp I tried to make the table names as static as I could, the name is only set by using `database.New(tablename)` and the string itself is unexported so we don't have to worry about it changing on us.

Maybe I'm misunderstanding where the driver comes into play, but I'm at least referencing that the database is postgres:

``` go
postgres, err := sql.Open("postgres", postgresOptions)
```

I think we should be OK without sanitizing the other fields.  The columns are constants so we can control them.  The table name is currently meant to be a hardcoded in whatever wrapper class we use (see [users/users.go](https://github.com/lighthouse/lighthouse/blob/provider-adding/users/users.go#L34) for how I defined the table name).  If we get to a point where we would like variable table names then we should sanitize it, but at this point we're OK.

@ngmiller @chfogerty @zach-taylor any thoughts sanitizing?

nickmonad:
Seems like a valid approach. As long as the non-sanitized strings used in the query build are not from user input, we should be ok.

Also, looks like we're currently pulling the deprecated pq package
https://github.com/lighthouse/lighthouse/blob/master/Dockerfile#L24

cmbrose:
@ngmiller the [bmizerany/pq reference](https://github.com/lighthouse/lighthouse/blob/master/Dockerfile#L24) import is where our postgres driver is coming from so we at least need to keep it for that (we get a runtime panic without it).

I did however move the reference from auth.go to postgres.go in 85de045 - seems like a better location for it.

@theMagicalKarp you found the package, is there something newer we use instead?

nickmonad:
@cmbrose Ah gotcha. I meant to say that it's deprecated in the sense that the main development has been moved to a new repository. https://github.com/bmizerany/pq --> https://github.com/lib/pq

cmbrose:
@ngmiller got it.  Changed and verified that everything still works: 87dfa81

theMagicalKarp:
+1 @ngmiller @cmbrose That was my mistake early on. I actually just realized that it was outdated a while ago, that's why I listed the newer one [earlier](https://github.com/lighthouse/lighthouse/pull/21#issuecomment-64046474) ^_^

nickmonad:
Ah sorry I missed that comment Rob :) All this looks pretty good. +1

nickmonad:
@cmbrose Before I forget, would you mind updating the API docs as well? Thanks!

cmbrose:
@ngmiller I've added the `/users/permissions/vms` endpoint.

@theMagicalKarp on that note, I think that the VM API is outdated, can you take a look at it?

cmbrose:
@chfogerty any input here?

nickmonad:
@cmbrose Is this good to merge? If we get this in, it will provide the basis for the /user endpoint in #22

cmbrose:
@ngmiller I've been waiting for @chfogerty to review this. I'll merge it by the end of the day though. 

@chfogerty you have been warned.

chfogerty:
+1

PR #22, opened 2014-12-01
Fix Routing to Handle Endpoints Persisted by Client

theMagicalKarp:
## What
- If the user is unauthorized they will be redirected to `/login`.
- If a login fails a `401` status code is returned.
- If a url persisted by the client is found & and the user is authenticated serve `index.html`.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -134,6 +134,12 @@ func Handle(r *mux.Router) {
 
         session.Save("auth", r, w)
 
+        if userOK && passwordOK {
+            w.WriteHeader(200)
+        } else {
+            w.WriteHeader(401)
+        }
+
         fmt.Fprintf(w, "%t", userOK && passwordOK)
     }).Methods("POST")
 
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -92,10 +92,10 @@ func AuthMiddleware(h http.Handler, ignorePaths []string) http.Handler {
 
         if loggedIn := session.GetValueOrDefault(r, "auth", "logged_in", false).(bool); loggedIn {
             h.ServeHTTP(w, r)
-        } else {
-            w.WriteHeader(http.StatusUnauthorized)
-            fmt.Fprintf(w, "unauthorized")
+            return
         }
+
+        http.Redirect(w, r, "/login",  http.StatusFound)
     })
 }
 
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -32,14 +32,18 @@ const (
     API_VERSION_0_1 = "/api/v0.1"
 )
 
+func ServeIndex(w http.ResponseWriter, r *http.Request) {
+    http.ServeFile(w, r, "static/index.html")
+}
+
+
 func main() {
 
     logging.Info("Starting...")
     baseRouter := mux.NewRouter()
 
-    baseRouter.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
-        http.ServeFile(w, r, "static/index.html")
-    }).Methods("GET")
+    baseRouter.HandleFunc("/", ServeIndex).Methods("GET")
+    baseRouter.NotFoundHandler =  http.HandlerFunc(ServeIndex)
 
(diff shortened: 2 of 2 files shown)


theMagicalKarp:
@lighthouse/owners

auth/auth.go
```diff
@@ -92,10 +92,10 @@ func AuthMiddleware(h http.Handler, ignorePaths []string) http.Handler {
 
         if loggedIn := session.GetValueOrDefault(r, "auth", "logged_in", false).(bool); loggedIn {
             h.ServeHTTP(w, r)
-        } else {
-            w.WriteHeader(http.StatusUnauthorized)
-            fmt.Fprintf(w, "unauthorized")
+            return
         }
+
+        http.Redirect(w, r, "/login",  http.StatusFound)
```
cmbrose, line 98: This redirect is giving me a 404 because nothing is set up to handle it.  If I prepend a `/api/v0.1` I still get a 404 because the redirect still sends a GET request (and `/login` is a POST).  Should we instead write a redirect to the response and let the frontend handle it?

robertsheehy-wf, line 98: @cmbrose do you have the latest static build from the lighthouse-client? It's probably reading `404` because `static/index.html` can't be found.

nickmonad:
Looks good. Client side updates here: https://github.com/lighthouse/lighthouse-client/pull/12

There still remains the issue of successful route/view changes on a manual route modification. (i.e. not via a link click) If I remove the isLoggedIn() check from the route change event listener and update the path like [this](https://github.com/lighthouse/lighthouse-client/pull/12/files#diff-8a5ab6578a797164084e00b78d5c78b6R12), these changes work fine and the routing responds accordingly. But, from what I can tell, the manual route change triggers some kind of refresh that invalidates the global user object on the client, causing the logged in check to fail and thus force a redirect to /login

I think we can persist state across manual route changes by having the backend return index.html with a user object pre-populated on the browser window object if the request is authenticated. Then, the front-end check for authentication will pass by checking that object and the routing will take over. The downside is it breaks the decoupling we have between the backend and frontend. The backend would be responsible for hosting index.html and pre-populating the user object via a template engine. :sob: 

All in all, I'm not really sure what to do about that. We can discuss more in our upcoming meeting and it's not really necessary now but something to keep in mind. Whatever needs to be done can happen in a future PR

+1

nickmonad:
(It's late so I'm kind of talking out my ass in that previous post.)

Now that I think about it, having the backend pre-populate the index.html wouldn't be so bad. The file can still live in lighthouse-client, but just needs to be written in such a way that supports whatever template engine would be on the backend.

robertsheehy-wf:
@ngmiller It's all good dude. Yeah I'm more use to having the server persist page states as opposed to Angular, so this is kinda new territory for me as well.
I'm still kinda against server side templating at this point since it seems to go against the idea behind Angular.
I'm in favor of a  `/api/v0.1/user` endpoint that would return information about the current user which could detail whether or not the user is logged in.  Depending on the results from that call on-page-load we can choose where we want to route the client.

I'm definitely down for discussing this later too.

cmbrose:
@ngmiller @theMagicalKarp a `/api/v0.1/user` would be easy to set up, https://github.com/lighthouse/lighthouse/pull/21 already has a `/api/v0.1/users/permissisons` endpoint group, so it should be easy to just turn `users` -> `user` and then condense the more general endpoints into it.

That being said, does `/api/v0.1/user` really _need_ to say if they're logged in or not?  If no user is logged in we won't have anything to return about them so could we just interpret no data as not logged in?

But that's for another PR

+1

auth/auth.go
cmbrose, line 98: @theMagicalKarp good call on that.  I probably should have been more on top of that :innocent:

nickmonad:
I think the /api/v0.1/user endpoint is worth a shot. It would essentially be like any other endpoint (returning 200 for a logged in user perhaps with a username and email, 401 otherwise) but would be semantically correct from an API perspective.

@theMagicalKarp I think this is good to merge

PR #27, opened 2015-01-27
Fixed query param forwarding

cmbrose:
This should solve the issue of query parameter forwarding over Docker requests.

Unfortunately mux doesn't make them very available like it does with its URL parameters, so I had to dig into the actual request a little.

Added a unit test to make up for it :)


cmbrose:
@lighthouse/owners 

@ngmiller hopefully this should solve your issue

nickmonad:
+1

cmbrose:
@lighthouse/owners I'm going to merge this later tonight if no one has any issues.

PR #28, opened 2015-01-28
Initial beacon interface

cmbrose:
### Major package changes
- **permissions**: Completely gone!  The concept of permissions has been moved into the beacons package and most security is moved to the beacon itself
- **aliases**: Moved from providers to beacons, otherwise unchanged.  This should open up providers to whatever changes/removals need to happen to it
- **beacons**: The primary new package.  Handles all administration, routing, and permissions to beacons
### The beacon model

Each Docker instance we know about is stored in a database paired with the following data:
- **BeaconAddress**: The IP address of the beacon hiding the instance
- **Token**: The security token of the beacon
- **Users**: A list of users who can access the beacon's token.
### How routing works
1. The frontend sends requests in the usual way:  `http://{LIGHTHOUSE}/d/{HOST}/{DOCKER ENDPOINT}` (Note the flip of the `/{HOST}` and `/d`)
2. Lighthouse takes the host parameter and searches it's database of known instances
   1. If the host is not known, assume that it is a pure instance and not a beacon.  The request is forwarded directly to the instance as `http://{HOST}/{DOCKER ENDPOINT}`
   2. If the host is known, extract the beacon's address. The request is forwarded to the beacon as `http://{BEACON}/d/{HOST}/{DOCKER ENDPOINT}`
      1. If the user is permitted to access the token, it is added to the request's header.
      2. Otherwise a default token (empty string) is added to the header.
3. ???
4. Profit
### Beacon control endpoints

**Note**: all endpoints are prefixed by `/api/v0.2` (not `v0.1` anymore!!)

Add a new Docker instance:

``` http
POST /beacons/create HTTP/1.1
Content-Type: application/json

{
    "Address": "<Beacon address>",
    "Token": "<Beacon token>",
    "Users": ["email1@example.com", "..."]
}
```

Permit a user to access a beacon's token:

``` http
PUT /beacons/user/{Instance}/{Id} HTTP/1.1
```

Revoke a users's permission to access a beacon's token:

``` http
DELETE /beacons/user/{Instance}/{Id} HTTP/1.1
```

Update a beacon's token:

``` http
PUT /beacons/token/{Instance} HTTP/1.1
Content-Type: application/json

"<Beacon Token>"
```

Update a beacon's address:

``` http
PUT /beacons/address/{Instance}/{Address} HTTP/1.1
```


cmbrose:
@lighthouse/owners for your review

nickmonad:
> If the host is not known, assume that it is a pure instance and not a beacon. The request is forwarded directly to the instance as http://{HOST}/{DOCKER ENDPOINT}

Do these semantics handle host aliases still? For instance, I should be able to address my local boot2docker instance as that and not via a raw IP

EDIT: I think the existence of `config/aliases.json` answers my question :)

beacons/beacons.go
```diff
@@ -0,0 +1,223 @@
...
+
+    updateRouter := r.PathPrefix("/update").Subrouter()
+
+    updateRouter.HandleFunc("/user", func(w http.ResponseWriter, r *http.Request) {
+        handleUpdateRoutes(w, r, AddUserToBeacon)
+    }).Methods("PUT")
+
+    updateRouter.HandleFunc("/user/rm", func(w http.ResponseWriter, r *http.Request) {
+        handleUpdateRoutes(w, r, RemoveUserFromBeacon)
+    }).Methods("PUT")
```
nickmonad: I think this call should instead be `DELETE /api/v0.2/beacons/user/{user email or id}`

nickmonad: Oh but you need the beacon address.... So maybe: `DELETE /api/.../beacons/{host}/user/{email}` ?

cmbrose:
@ngmiller Yup!  Aliases are figured out before the Docker instance lookup so using them in requests is still fine.

That does bring up that when you use the `/beacons/create` endpoint, that instance address can't be aliased right now.  Adding an alias lookup to that endpoint is easy though, so I can add that if it would be helpful. (Which I would think it would be)

nickmonad:
@cmbrose I agree on the alias lookup for /beacons/create

beacons/beacons.go
cmbrose: Yeah, I wasn't really sure what the most consistent way to do that was without the request body in the DELETE, but I like your way better.

For consistency then, I could format all of these update endpoints to look like
`METHOD /api/.../beacons/{host}/{field to update}/{data}`

So a user remove would be `DELETE /api/.../beacons/{host}/user/{email}`
and a token update would be `PUT /api/.../beacons/{host}/token/{token}`

cmbrose: Would that lead to any security issues though, putting tokens and emails and addresses right into the URL?

nickmonad:
Why does the /beacons/create object sent over need a docker instance address? It's beacon's job to discover those right?

beacons/beacons.go
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        err = json.Unmarshal(reqBody, &body)
+        if err != nil {
+            w.WriteHeader(http.StatusInternalServerError) 
+            fmt.Fprint(w, err)
+        }
+
+        if body.InstanceAddress == "" {
+            w.WriteHeader(http.StatusInternalServerError)
```
nickmonad: These should probably be a 400 https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_Error

cmbrose:
Yeah, you're right. I sort of forgot the beacon had a `/vms` endpoint to search from... ;)

beacons/beacons.go
nickmonad: Yeah that's a good point. Should probably adhere to REST "best" practices for most things and make an exception case for tokens

beacons/beacons.go
cmbrose: I'll update to a [400](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#400)

beacons/beacons.go
nickmonad: As for user emails, we could generate a unique id that's associated with that particular user

cmbrose: We can look into that for users, but it'll probably need to be a different branch because there's a lot of databases that would be affected by that.  I would imagine that for our own testing passing emails is fine though. (Seeing as our only user is a junk user anyways)

nickmonad: :+1:

cmbrose:
@ngmiller see updated endpoints

I didn't add docker instance aliasing because I just ask the beacon for addresses now instead.  It's on the user to set up any aliases they want for the frontend at this point.

@lighthouse/owners for the new endpoints, what sort of return values do you want/need?  At this point I'm just reporting success or failure with no actual data.

cmbrose:
Unfortunately full endpoint testing is really gross with our security :( so I haven't done it yet

nickmonad:
For the sake of resisting future naming headache, it seems like the `/beacons/{Instance}/<noun>` calls should be flipped to `/beacons/<noun>/{Instance}`. With the current setup, we'd have to prevent users from using "create" as an instance name as that would conflict with the `/create` endpoint.

cmbrose:
@ngmiller that also makes the request more readable I think.  I flipped them around (see updated endpoints above)

I also realized that I wasn't aliasing the update endpoints, so I've added that feature as well.

nickmonad:
Looks good +1

cmbrose:
@lighthouse/owners - unless anyone has requests to change the new endpoints, this is ready to go!

chfogerty:
+1

cmbrose:
@theMagicalKarp speak now or forever hold your peace

theMagicalKarp:
+1

PR #29, opened 2015-01-30
Fixed database concurrency issue

cmbrose:
This solves a concurrency issue with lazily loading databases.  In short, all database must now be loaded at boot.

### DIFF
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -20,7 +20,7 @@ import (
 
     "github.com/lighthouse/lighthouse/auth"
     "github.com/lighthouse/lighthouse/provider"
-    "github.com/lighthouse/lighthouse/providers/aliases"
+    "github.com/lighthouse/lighthouse/provider/aliases"
     "github.com/lighthouse/lighthouse/handlers"
     "github.com/lighthouse/lighthouse/users"
     "github.com/lighthouse/lighthouse/users/permissions"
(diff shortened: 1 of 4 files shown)


cmbrose:
@ngmiller can you verify that this fixes your issue?  I still can't run that darn frontend :)

provider/aliases/aliases.go
```diff
@@ -28,13 +29,19 @@ var Aliases *databases.Table
 
 func getDBSingleton() *databases.Table {
     if Aliases == nil {
+        panic()
```
cmbrose: The panics are only to help us remember to load these at boot.  If Init() has been called we should have no troubles.

lighthouse.go
```diff
@@ -20,9 +20,11 @@ import (
 
     "github.com/lighthouse/lighthouse/auth"
     "github.com/lighthouse/lighthouse/provider"
+    "github.com/lighthouse/lighthouse/providers/aliases"
```
nickmonad: `/providers/` --> `/provider/`

provider/aliases/aliases.go
nickmonad: I got some errors upon startup about a lack of arguments to panic, which I fixed with simple strings - but I'm not sure if that's correct

nickmonad:
A few notes, but otherwise it works! +1

cmbrose:
My bad on that! Sort of scary that Travis misses stuff like that...

PR #32, opened 2015-02-07
Migrate to goji

cmbrose:
**Note:** this is a PoC.  This isn't complete (and may never be), but this is one potential solution to the parameter issue currently going on.

The first (and most obvious) change is a migration from gorilla/mux to goji.  This was originally in a hope that goji would automatically fix the problem, but it did not.

It did, however, lend some interesting tools that can help us out here. Namely easy middleware and easy global data.  I created a short middleware function (QueryParamExtract in lighthouse.go) which takes the URI encoded request and extracts the pieces of it into the map of global data.

For example `/beacons/user/localhost%2Fv1.12/username` would be parsed into 

```
{
    0 : "beacons",
    1 : "user",
    2 : "localhost/v1.12",
    3 : "username"
}
```

This will fix the routing issues, but of course has tradeoffs
1. We lose precision in our endpoints. Instead of saying `/beacons/user/{Instance}/{Id}`, we can now only be exact as `/beacons/user/*`
2. We have to error check our queries. Someone could access `/beacons/user/boot2docker` and would hit `/beacons/user/*`.  The fix is simple length checking.
3. We can only use "leaf" endpoints. If we want an endpoint of `/beacons/user/*`, we cannot also have `/beacons/*` because these would interfere.


cmbrose:
@lighthouse/owners do you guys have any thoughts on pursuing this or not?

cmbrose:
It also occurs to me that this exact same idea could be done in the current system using the `sessions` library we already have.  So if you guys are opposed to switching to goji I can work that up as well.

cmbrose:
@lighthouse/owners this is ready for review.

A note to anyone who has already gotten goji for beacon - you may want to pull it down again because they made a breaking change 9 days ago...

lighthouse.go
```diff
@@ -38,34 +41,52 @@ func ServeIndex(w http.ResponseWriter, r *http.Request) {
...
+    baseRouter.Use(logging.Middleware)
 
-    baseRouter.HandleFunc("/", ServeIndex).Methods("GET")
-    baseRouter.NotFoundHandler =  http.HandlerFunc(ServeIndex)
+    baseRouter.NotFound(ServeIndex)
 
     staticServer := http.FileServer(http.Dir("static"))
-    baseRouter.PathPrefix("/static/").Handler(
-        http.StripPrefix("/static/", staticServer))
+    baseRouter.Get("/static/*", staticServer)
```
nickmonad, line 79: Having trouble getting static resources to load. `index.html` loads but subsequent requests for scripts and css files return a 404

nickmonad, line 79: Re-adding `http.StripPrefix("/static/", staticServer)` seems to work

nickmonad:
Just the one issue with static resources. Otherwise looks good! +1

cmbrose:
@lighthouse/owners after discussing with @ngmiller a little bit, it might be best to just drop this switch. Goji has some pretty big flaws in endpoint regexes and backwards compatibility.  Everything I added here for solving the encoding issue can be done in gorilla/mux.  Unless anyone is opposed to not switching, I'm going to close this later today.

cmbrose:
Closing as https://github.com/lighthouse/lighthouse/pull/35 is a better solution.

PR #33, opened 2015-02-09
Applications Database/Rollback Groundwork

chfogerty:
This is for adding functionality for to keep track of applications and containers within Lighthouse, as noted by [this](https://trello.com/c/UYcyoRKe) Trello card. Database design is shown below.

Lighthouse database

I had to add some functionality to the databases package in order to allow for a schematized table, as I'm going to need filtering functionality for history and deployment rollbacks. This needs to be cleaned up a bit, so feel free to make recommendations on that.

The other specific issue I'm looking for feedback on is handlers/endpoints (for containers and applications). Not really sure how to structure these, or what exactly we'll want available.

### DIFF
--- a/containers/applications/applications.go
+++ b/containers/applications/applications.go
@@ -1,0 +1,73 @@
+// Copyright 2014 Caleb Brose, Chris Fogerty, Rob Sheehy, Zach Taylor, Nick Miller
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package applications
+
+import (
+    "fmt"
+    "errors"
+
+    "github.com/lighthouse/lighthouse/databases"
+    "github.com/lighthouse/lighthouse/databases/postgres"
+)
+
+var applications *databases.Table
+
+var columns = []string {"Id", "Name"}
+var types = []string {
+    "serial primary key",
+    "text",
+}
+
+type Application struct {
+    Id int
+    Name string
+}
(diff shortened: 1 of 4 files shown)


chfogerty:
@lighthouse/owners Questions, comments, concerns?

containers/applications/applications.go
```diff
@@ -0,0 +1,73 @@
...
+    "github.com/lighthouse/lighthouse/databases/postgres"
+)
+
+var applications *databases.Table
+
+var columns = []string {"Id", "Name"}
+var types = []string {
+    "serial primary key",
+    "text",
+}
```
cmbrose, line 31: Instead of using parallel arrays, could we maybe use maps?  Just adding the type

``` go
type Schema map[string]string
```

would let you turn this into

``` go
var schema = databases.Schema {
    "Id" : "serial primary key",
    "Name" : "text",
}
```

Which I think is much more readable (and helps make sure types/names are aligned)

databases/databases.go
```diff
@@ -81,6 +106,36 @@ func (this *Table) Insert(key string, value interface{}) error {
     return err
 }
 
+func (this *Table) InsertSchema(columns, values []string) error {
+    var colBuf, valBuf bytes.Buffer
+    queryVals := make([]interface{}, len(values))
+
+    if len(columns) != len(values) {
+        return errors.New(`Column/Value mismatch`)
+    }
```
cmbrose, line 115: Using maps would also get rid of the need to length-check the arrays like you do here.

databases/databases.go
```diff
@@ -113,3 +168,15 @@ func (this *Table) SelectRow(key string, value interface{}) error {
 
     return err
 }
+
+func (this *Table) CustomSelect(query string, queryParams []string) (row *sql.Row) {
```
cmbrose, line 172: The only issue I have with letting the user give their own queries is backwards-compatibility as we add new database types.  It should be pretty easy to just take a map of `{ col1 : val1, col2 : val2, ... }` and create a standard query out of it.

The only tricky part would be sticking the return value into an `interface{}`, but even that would be the same as the old stuff I would think.

cmbrose:
@chfogerty I really like the new schema stuff.  Left a few comments above.

chfogerty:
Closed - splitting into 2 PRs to push schema update through faster.

Issue #30, opened 2015-02-04
URI encoded parameters don't work with gorilla/mux


cmbrose:
For reference see [the issue in gorrila/mux](https://github.com/gorilla/mux/issues/77)

Even with encoded requests, mux _still_ splits the parameters on `/`s.  For right now we can continue using aliases to get around it, just don't make aliases with forward slashes in them ;)

@chfogerty @theMagicalKarp do you have opinions/thoughts on solutions?  It would be nice to not have to drop mux this late in the game, but this is pretty big.

chfogerty:
The current workaround is to use query parameters.

nickmonad:
For the sake of transparency and as per Chris, Caleb, and I's discussion, I think sticking major routing parameters in the query params should be a last resort

nickmonad:
Limiting the characters used in host aliases would be another option. Client side validation would be straightforward (but of course would have to be checked again server side)

cmbrose:
@theMagicalKarp for beacon you used a different routing library.  Do you think that's something we could switch to if we need to?

I'm not a fan of relying purely on aliasing for this, but I agree with Nick too that query parameters isn't a great thing either.

robertsheehy-wf:
Are you guys talking about where slashes for the docker request are encoded out?

```
some.beacon.ip/d/some.vm.ip/v1.4/_ping
to
some.beacon.ip/d/some.vm.ip%2Fv1.4%2F_ping
```

Because both are valid in beacon. Right now I'm using [goji](https://goji.io/) and I :heart: it.

Not that I'm saying we should change the standard now, but this is kinda why I liked having every request be a `POST` detailing how lighthouse should make the request.

```
POST /docker
{
    "method": "GET",
    "url": "/v1.4/_ping",
    "vm": "127.0.0.1:2375",
    "body": {"someAwesome": "json", "or": "query params"},
    "metadata": "extra stuff that lighthouse can use"
}
```

cmbrose:
@robertsheehy-wf yeah, with gorilla/mux the encoding is resolved before the routing for some reason so there's basically no point to it...

How tough would it be to move to goji do you think?

robertsheehy-wf:
It's hard to say.  We have a lot of code that's coupled with gorilla. The pattern of ...

``` Go
func handler(w http.ResponseWriter, r *http.Request) {
```

is pretty consistent among Go web frameworks though.

cmbrose:
[...]

PR #35, opened 2015-02-11
Generic URI extraction

cmbrose:
## What

Fix for [the URI encoding issue](https://github.com/lighthouse/lighthouse/issues/30).
## Why

So that URL parameters such as hosts can be URI encoded and captured correctly.
## How

There are a few endpoint changes to support it (in code only, the API remains unchanged).  The focus of the fix is the addition of the `GetEndpointParams()` function in `handlers/handlers.go`.

---

The rest of this comment is simply documentation on this function for future use.

`GetEndpointParams()` takes an `*http.Request` and `[]string` and returns a `map[string]string`.  The values of the array are used as keys in the map and parallel the endpoint itself.

To use this function, mux must have a `"Endpoint"` route variable (`r.HandleFunc("/prefix/{Endpoint:.*}", ...`)

The endpoint is broken by `/`s and the map is filled until either the array or the endpoint runs out of new values.  If the array runs out first, the rest of the endpoint is placed into the last key in the array.  If the endpoint runs out first, the remaining keys in the array are ignore (this lets you perform size checks).
### Examples

**Normal case**

``` json
Endpoint:  /d/127.0.0.1%2Fv1.12/info
Key array: [ "Host", "Endpoint" ]
Result:    { "Host" : "127.0.0.1/v1.12", "Endpoint" : "info" }
```

**Not enough keys**

``` json
Endpoint:  /d/127.0.0.1%2Fv1.12/containers/json
Key array: [ "Host", "Endpoint" ]
Result:    { "Host" : "127.0.0.1/v1.12", "Endpoint" : "containers/json" }
```

**Too many keys**

``` json
Endpoint:  /d/127.0.0.1%2Fv1.12/info
Key array: [ "Host", "Endpoint", "ExtraKey" ]
Result:    { "Host" : "127.0.0.1/v1.12", "Endpoint" : "info" }
```

### DIFF
--- a/beacons/beacons.go
+++ b/beacons/beacons.go
@@ -26,8 +26,8 @@ import (
 
     beaconStructs "github.com/lighthouse/beacon/structs"
 
+    "github.com/lighthouse/lighthouse/handlers"
     "github.com/lighthouse/lighthouse/beacons/aliases"
-
     "github.com/lighthouse/lighthouse/databases"
     "github.com/lighthouse/lighthouse/databases/postgres"
 )
@@ -37,6 +37,10 @@ const (
 )
 
 var beacons *databases.Table
+
+var (
+    NotEnoughParametersError = errors.New("beacons: not enough parameters given")
+)
 
 type Beacon struct {
     Address string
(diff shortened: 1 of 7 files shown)


cmbrose:
@lighthouse/owners for your review

chfogerty:
I still need to pull this down and test it, but it looks good.

robertsheehy-wf:
+1

nickmonad:
Works locally! +1

chfogerty:
Well then. +1

PR #34, opened 2015-02-09
Remove "provider" module because of Beacon

robertsheehy-wf:
[trello#71](https://trello.com/c/UnIzWMPz/71-remove-beacon-functionality-components-from-lighthouse)

PR #36, opened 2015-02-12
Beacons/Instance Listing and Initial Database Schema Support

cmbrose:
## What

Primarily (maybe not in terms of number of added lines) the ability to request all known beacons and instances behind those beacons.
## How

Most of the work here was adding schemas to the `database` package.  Note that these updates _are not complete_ and only implemented enough to support the new endpoints.  A later PR  will clean `databases` and eventually we will be changing to **only** schema-ed model (currently we support schemas and simple key-value mappings).

The actual endpoints use the new database models to more efficiently extract address and other information from the database.  Their implementations are straightforward.
## Testing

A lot of this PR also includes large testing improvement capabilities.  Previously we could only mock out a database connection which left us _very_ limited in terms of what we could feasibly test.  The file `databases/mocktable.go` solves this issue by providing an interface for a completely custom database which doesn't require any external services.  Future PRs will probably be needed to go through other packages to make use of this to improve our unit tests.
### Endpoints

Get a **distinct** list of beacons which can be accessed by the current user

``` http
GET /beacons/list HTTP/1.1
```

Get a **distinct** list of instances behind a particular beacon which can be accessed by the current user
**Note:** Once schemas become widespread, this list will contain the instance's aliases instead of address where applicable

``` http
GET /beacons/list/{BeaconAddress}
```
### Examples

Request:

``` http
GET /beacons/list HTTP/1.1
```

Response:

``` http
HTTP/1.1 200 OK
[ "127.0.0.1:5002" ]
```

Request:

``` http
GET /beacons/list/127.0.0.1:5002 HTTP/1.1
```

Response:

``` http
HTTP/1.1 200 OK
[ "127.0.0.1:5001/v1.12" ]
```


cmbrose:
@lighthouse/owners for your reading pleasure

cmbrose:
Also, once this is closed, we can also close https://github.com/lighthouse/lighthouse/pull/34

chfogerty:
+1. With this, My stuff for this sprint should be pretty easy. I can start cleaning up the code here, then, as well as improving the scan tests.

beacons/handlers.go
```diff
@@ -0,0 +1,190 @@
...
+    if err != nil {
+        return http.StatusInternalServerError, err
+    }
+
+    beacon := beaconData{"", beaconInfo.Address, beaconInfo.Token, userMap{}}
+    for _, user := range beaconInfo.Users {
+        beacon.Users[user] = true
+    }
+
+    vmsTarget := fmt.Sprintf("http://%s/vms", beacon.BeaconAddress)
```
robertsheehy-wf, line 161: Most of the time beacon is going to be hosting a [TLS connection](https://github.com/lighthouse/beacon/blob/master/beacon.go#L127) with self signed certs to encrypt traffic. So the server will need to generate ssl certs & keys for secure communication between beacon and lighthouse.  To generate keys you can run something like.

```
openssl req -new -nodes -x509 -out client.pem -keyout client.key -days 3650 -subj "/O=Lighthouse/OU=IT/CN=146.148.80.171"
```

I have a basic implementation of a beacon client [here](https://gist.github.com/theMagicalKarp/5d0cf8762e8cdecde784) if that helps at all.

beacons/beacons.go
```diff
@@ -90,34 +112,23 @@ func LoadBeacons() map[string]Beacon {
...
 
     r.HandleFunc("/user/{Endpoint:.*}", handleAddUserToBeacon).Methods("PUT")
 
     r.HandleFunc("/user/{Endpoint:.*}", handleRemoveUserFromBeacon).Methods("DELETE")
 
-    r.HandleFunc("/address/{Endpoint:.*}", handleUpdateBeaconAddress).Methods("PUT")
-
     r.HandleFunc("/token/{Endpoint:.*}", handleUpdateBeaconToken).Methods("PUT")
 
     r.HandleFunc("/create", func(w http.ResponseWriter, r *http.Request) {
```
nickmonad: Should this handler function be moved to its own `handleCreateBeacon`?

beacons/beacons.go
```diff
@@ -129,189 +140,43 @@ func Handle(r *mux.Router) {
...
+            fmt.Fprint(w, string(output))
+        }
 
-    defer resp.Body.Close()
+    }).Methods("GET")
 
-    if resp.StatusCode != http.StatusOK {
-        return resp.StatusCode, errors.New("beacon error")
-    }
+    r.HandleFunc("/list/{Beacon:.*}", func(w http.ResponseWriter, r *http.Request) {
```
nickmonad: Same here. Should this be a named function that's passed in instead of inlined here? I'm fine with inlining the handlers, but we should be consistent. This also separates the endpoint url from its HTTP method by a fair amount.

databases/databases_test.go
```diff
@@ -81,8 +107,75 @@ func Test_Insert(t *testing.T) {
...
+    midCol := revData[args[1]] != firstCol &&
+              revData[args[1]] != lastCol &&
+              revData[args[1]] != ""
+
+    assert.True(t, midCol,
+        fmt.Sprintf("Middle query param %v is incorrect", args[1]))
+}
+
+func min(locs map[string]int, a, b string) string {
+    if locs[a] < locs[b] {
```
nickmonad: Should this be `loc[a] <= loc[b]` to give preference to a? Or does it not matter?

chfogerty: It doesn't matter for the current use, and the function isn't exported. Current use is comparing the location of substrings, and it should probably be rewritten to express that more clearly. I'll fix it after class today.

beacons/handlers.go
cmbrose, line 161: Are we going to need to hold onto these for every time we talk to a beacon then?

beacons/beacons.go
cmbrose: Yeah, I'll move this and the next handler to `beacons/handlers.go`

beacons/handlers.go
cmbrose, line 161: Also, how will the the backend actually get these?  Will the frontend be sending them or will I need to generate them on the spot?

beacons/handlers.go
```diff
@@ -0,0 +1,239 @@
...
+
+    err = json.Unmarshal(reqBody, &beaconInfo)
+    if err != nil {
+        return
+    }
+
+    beacon := beaconData{"", beaconInfo.Address, beaconInfo.Token, userMap{}}
+
+    currentUser := session.GetValueOrDefault(r, "auth", "email", "").(string)
+    beacon.Users[currentUser] = true
```
cmbrose, line 155: @ngmiller @zach-taylor the current user is automatically given privileges to the beacon they're adding, so you won't need to include them in the list of pre-approved users.

zach-taylor:
@cmbrose REST endpoints typically have /{resource}s as the list endpoint.

So in this case it would be: `GET /beacons HTTP/1.1`.

Otherwise this looks good.

beacons/handlers.go
robertsheehy-wf, line 161: Generally the client key & cert should be generated before lighthouse is ran.  I'm pretty sure lighthouse should only need one key & cert for communicating to several beacons.

cmbrose, line 161: Created a ticket for a later PR [here](https://trello.com/c/2VrJotdH/100-update-to-tls-clients)

cmbrose:
@zach-taylor do you mean in the actual API definition?  I can change that once I update the real API for this change.

cmbrose:
@lighthouse/owners any other comments?  I think I've addressed everything so far.

zach-taylor:
@cmbrose `/beacons` for the list endpoint and `/beacons/{BeaconAddress}` for the individual beacon endpoint rather than `/beacons/list` and `/beacons/list/{BeaconAddress}`

If that's not too much trouble

cmbrose:
I can make `/beacons` list beacons, but the `/beacons/{BeaconAddress}` has to be wild-carded so that the address could contain `/`s.  That would fight the existing endpoints (`/user` for example) because mux wouldn't be able to tell whether or not `/user` was the endpoint you wanted, or part of the address for listing.

tl;dr: I can make `/beacons` happen, but not `/beacons/{BeaconAddress}`.  For consistency I think it should stay how it is.

zach-taylor:
@cmbrose ah I see. fair enough.

robertsheehy-wf:
+1

nickmonad:
+1

PR #37, opened 2015-02-18
User actions

cmbrose:
**Work in progress**
## What
- User control endpoints
- Better authorization of users/resources
## Why
- Can add/edit users
- Can create a user permission hierarchy
- Can transfer ownership/privileges of resources
## How
- The primary additions are `auth/users.go` and `auth/permissions.go` which contain the User and Permission types.
- Users are given an `AuthLevel` field which specifies what other users they can control.
  - Level is uncapped.  Default is 0.  At level 1+ you can create new users (who are level 0)
  - If your `AuthLevel` is >= another user's, you can view and edit them
  - You cannot modify anyone's `AuthLevel` to be higher than your own (including yourself of course)
- Users are also given a `Permissions` field which specifies their control over resources (like Beacons)
  - Level 0 - access resource (e.g. get token for Beacon)
  - Level 1 - modify resource (e.g. add/remove users from a Beacon or change token)
  - Level 2 - own resource.  Modifiers can't remove you (unless they are also an owner)
## Endpoints

**Note** since beacon permissions are now paired with Users, the `/beacon/user` endpoints are gone.

Get a list of User you can view (list includes current User)

``` http
GET /users/list HTTP/1.1
```

Get information about a User (must be able to view them)

``` http
GET /users/{Email} HTTP/1.1
```

Modify a User (must be able to modify them)

``` http
PUT /users/{Email} HTTP/1.1
{
    "AuthLevel" : <Level>,
    "Password" : <Password>,
    "Beacons" : {
        "Address 1" : <Level>,
        "Address 2" : <Level>
    }
}
```

Create a User (must have AuthLevel >= 1)

``` http
POST /create HTTP/1.1
{
    "Email" : <Email>,
    "Password" : <Password>
}
```
### Examples

**Get a list of viewable Users**

``` http
GET /users/list HTTP/1.1
```

Response:

``` http
HTTP/1.1 200 OK
[ "joe.schmoe@example.com", "john.doe@example.com" ]
```

**Get a User you can view**
Request:

``` http
GET /users/joe.schmoe@example.com HTTP/1.1
```

Response:

``` http
HTTP/1.1 200 OK

{
    "AuthLevel" : 0,
    "Permissions" : {
        "Beacons" : {
            "127.0.0.1:5000" : 0
        }
    }
}
```

**Get a User you _can't_ view**
Request:

``` http
GET /users/admin@example.com HTTP/1.1
```

Response:

``` http
HTTP/1.1 404 Not Found
```

**Update a User's password**
Request:

``` http
PUT /users/joe.schmoe@example.com
{
    "Password" : "i_LiKe_K1773N$"
}
```

Response:

``` http
HTTP/1.1 200 OK
```

**Update a User you are not allowed to modify**
Request:

``` http
PUT /users/admin@example.com
{
    "Password" : "WORST_BOSS_EVER"
}
```

Response:

``` http
HTTP/1.1 401 Unauthorized
```

**Remove a User from a resource**
Request:

``` http
PUT /users/joe.schmoe@example.com
{
    "Beacons" : {
        "Address 1" : -1,
    }
}
```

Response:

``` http
HTTP/1.1 200 OK
```

**Create a User**
Request:

``` http
POST /users/create
{
    "Email" : "joe.schmoe2@example.com",
    "Password" : "YUBsoR00d?"
}
```

Response:

``` http
HTTP/1.1 200 OK
```


cmbrose:
@lighthouse/owners this isn't ready to get put in yet, but looking for comments.

auth/users.go
```diff
@@ -0,0 +1,358 @@
...
+            return int(i)
+        case int32:
+            return int(i)
+        case int64:
+            return int(i)
+        case float32:
+            return int(i)
+        case float64:
+            return int(i)
+    }
```
nickmonad: Hmmmm - why not just `return int(i)` and call it good? Does `int(i)` return an error or throw an exception if it can't convert for some reason?

nickmonad: If the value to convert is coming from user input, we may want to look into a validation library. Something that could allow us to declaratively state rules like "this field is optional, this field is required, this must be an int, etc

I really like the way this python lib handles it: https://github.com/alecthomas/voluptuous

cmbrose: This is specifically used to extract AuthLevels for resources in Permissions.  The issue is that, for example, update requests for beacons come in as

``` json
"Beacons" : {
    "BeaconAddr1" : 1
}
```

The `"BeaconAddr1" : 1` is parsed by the `json` package into a `map[string]interface{}` (I have no control over that, that's the only way it will interpret this).  So the `string` is the beacon addresses and the `interface{}` is the AuthLevel.  The issue is that, even though we know these level are `int`s, `json` does not.  So it assigns it some numeric type, but not always an `int` (from my experience it's usually a `float64`).

So now we have this number, but don't know the type.  Since Go strictly enforces types, we have to do some sort of validation on it.  The cleanest way that I've found is this switch.  Just returning `int(i)` is a compile error because there's no type check, and returning `i.(int)` will cause a panic because even in the best case `i` is probably a `float64`, but could be anything really.

nickmonad: Damn. So yeah what about something like https://github.com/absoludity/goforms? (This specific lib might not be usable if it can't support nested objects, but the idea is there)

EDIT: found via a search for "golang data validation"

cmbrose: Did some digging into exactly json parses the data and found this in their documentation about [Unmarshal](http://golang.org/pkg/encoding/json/#Unmarshal):

```
bool, for JSON booleans
float64, for JSON numbers
string, for JSON strings
[]interface{}, for JSON arrays
map[string]interface{}, for JSON objects
nil for JSON null
```

So, trusting that, I can bring this down to only checking against float64 instead of literally every numeric type Go has!

auth/users.go
```diff
@@ -0,0 +1,338 @@
...
+func parseUserUpdateRequest(curUser, modUser *User, updateJSON []byte) (map[string]interface{}, int) {
+    
+    updates := struct {
+        AuthLevel int           `json:omitempty`
+        Password string         `json:omitempty`
+        Beacons map[string]int  `json:omitempty`
+    }{
+        AuthLevel: modUser.AuthLevel, 
+        Password: modUser.Password,
+    }
```
cmbrose, line 297: I realized that dealing with generic JSON for this method was just silly.  By defining this struct we can ensure correct types on the inputs.

auth/users.go
```diff
@@ -0,0 +1,338 @@
...
+func GetUser(email string) (*User, error) {
+    user := &User{}
+    where := databases.Filter{"Email" : email}
+    err := getDBSingleton().SelectRowSchema(nil, where, user)
+
+    if err != nil {
+        return nil, err
+    }
+
+    user.convertPermissionsFromDB()
```
cmbrose, line 101: The only current catch is that anytime Permissions are read from the database, we will run into the same issue with float vs int typing.  So this function goes through and casts everything to an int so types are consistent.

zach-taylor:
+1

cmbrose:
@chfogerty @theMagicalKarp @ngmiller for your review

chfogerty:
+1

nickmonad:
@cmbrose Looks good! My only concern is using 401 responses in light of the new auth semantics. Could we change them to 403 (Forbidden)? https://en.wikipedia.org/wiki/HTTP_403

cmbrose:
@ngmiller definitely.  I had meant to do that after our discussion about 401 vs 403 a couple days ago.

nickmonad:
With that, +1

robertsheehy-wf:
+1

auth/users.go
```diff
@@ -0,0 +1,336 @@
...
+
+    if err != nil {
+        writeResponse(w, http.StatusNotFound, UserAccessError) 
+        return
+    }
+
+    currentUser := GetCurrentUser(r)
+
+    if !currentUser.CanViewUser(reqUser) {
+        writeResponse(w, http.StatusNotFound, UserAccessError)
```
cmbrose, line 192: @ngmiller just to verify, 404's here won't mess with our auth-refresh fix right?

nickmonad, line 192: Nah I think we're good. The only exceptional status code we consider on the front end is 401 to redirect to /login. A 404 would should up as a standard alert

PR #38, opened 2015-02-22
Alias add/update endpoint

cmbrose:
Adds a new endpoint to manage aliasing:

``` http
PUT /aliases/{Alias} HTTP/1.1

"Address"
```

If the alias didn't already exist it will be created, otherwise the address is updated.

If the address is empty, returns status code `BadRequest (400)`.  Otherwise, unless there is some internal error, `OK (200)`.


nickmonad:
+1 with API doc updates

theMagicalKarp:
+1

PR #41, opened 2015-02-23
Application Container Groupings

chfogerty:
Essentially what #33  provided, but complete and with the updated database interface. That is:

> This is for adding functionality for to keep track of applications and containers within Lighthouse, as noted by [this](https://trello.com/c/UYcyoRKe) Trello card. Database design is shown below.
> 
> Lighthouse database
> 
> I had to add some functionality to the databases package in order to allow for a schematized table, as I'm going to need filtering functionality for history and deployment rollbacks. This needs to be cleaned up a bit, so feel free to make recommendations on that.
> 
> The other specific issue I'm looking for feedback on is handlers/endpoints (for containers and applications). Not really sure how to structure these, or what exactly we'll want available.


chfogerty:
@lighthouse/owners Check it out, see if there's anything you dislike.

cmbrose:
+1.  But we should update the other packages that use databases to the new `databases.Insert()` return values before merging.

containers/containers.go
```diff
@@ -0,0 +1,83 @@
...
+    "github.com/lighthouse/lighthouse/databases"
+    "github.com/lighthouse/lighthouse/databases/postgres"
+)
+
+var containers databases.TableInterface
+
+var schema = databases.Schema {
+    "Id" : "serial primary key",
+    "AppId" : "integer REFERENCES applications (Id)",
+    "DockerInstance" : "text REFERENCES aliases (keyColumn)",
```
cmbrose: Just making a note that this column name is changing with https://github.com/lighthouse/lighthouse/pull/38

chfogerty:
@cmbrose Only `databases.InsertSchema()` has a new return value, and that (as far as I can tell) has been updated. While you can still get the integer for `databases.Insert()`, the utility of that integer is significantly reduced, unless you're keying off some form of integer. From the [godocs](http://golang.org/src/database/sql/sql.go?s=44855:45361#L1719):

```
// LastInsertId returns the integer generated by the database
// in response to a command. Typically this will be from an
// "auto increment" column when inserting a new row. Not all
// databases support this feature, and the syntax of such
// statements varies.
```

cmbrose:
@chfogerty my mistake, I thought that schema-ed `aliases` and `beacons` were already in master.  The PRs to do that are out though, so we need to be careful of what order these all get merged in.

containers/handlers.go
```diff
@@ -0,0 +1,77 @@
...
+
+    if err != nil {
+        errInfo.StatusCode = http.StatusInternalServerError
+        errInfo.Cause = "Database read error"
+        errInfo.Message = "Failed to read application from database"
+        return errInfo
+    }
+    
+    //create container
+    containerId, err := CreateContainer(appId, info.Host, name)
```
chfogerty: @cmbrose Just to be 100% sure, is `info.Host` a Docker alias, Docker address, both, or neither?

cmbrose:
@chfogerty it will be the actual Docker address

cmbrose:
@chfogerty would it be possible to move the new `applications` and `containers` folders into the `handlers` folder? It would be nice to keep the total number of packages at the root lower, especially since only things in `handlers` will we calling these methods for the most part.

chfogerty:
@cmbrose I moved the package into handlers.
@lighthouse/owners Any other comments?

cmbrose:
+1 from me

nickmonad:
+1

theMagicalKarp:
+1

handlers/docker/docker.go
```diff
@@ -124,6 +126,7 @@ func DockerHandler(w http.ResponseWriter, r *http.Request) {
 
     var customHandlers = handlers.CustomHandlerMap{
         //regexp.MustCompile("example"): ExampleHandler,
+        regexp.MustCompile("/containers/create.*") : containers.ContainerCreateHandler,
```
cmbrose: @chfogerty I tracked down the issue @ngmiller was having with getting the handler to run.  The endpoint given in the handler info doesn't have a leading `/`, so this just needs to be `containers/create.*` instead.

chfogerty: Fixed.

chfogerty:
@lighthouse/owners Okay, this absolutely should work now that I did more than just the unit tests. Please take a last look over everything.

databases/databases_test.go
```diff
@@ -107,23 +107,63 @@ func Test_Insert(t *testing.T) {
         "Insertion call should encode given value as JSON")
 }
 
-func Test_InsertSchema(t *testing.T) {
+//TODO
+//Will not run because we can't mock sql.Row
+//We need to write a database driver...
+//Or use someone else's
+/*
+func Test_InsertSchema_WithReturn(t *testing.T) {
```
chfogerty, line 115: So it turns out I can't test anything that uses sql.Row. Which means this can't be tested. Left a TODO here to remind us to look into it if anyone gets some free cycles.

chfogerty:
@lighthouse/owners Last chance to say something. Have a 3 hour drive, then I'm merging this. I'm going to fix the tests in a different PR.

zach-taylor:
@chfogerty looks like there's merge conflicts that need to be resolved before you merge.

chfogerty:
@zach-taylor Yep. It's resolved locally, but I must have forgotten to push it up last night.

PR #43, opened 2015-02-23
Coveralls

cmbrose:
This is a request for input on how best to handle this.  I'm not very familiar with how Travis CI scripts work, but I have a rough one working for code coverage (see `.travis.yml`).

The script uses [Coveralls](https://coveralls.io/) to get the coverage from `go test`'s output.

Currently we have to specify each package separately because Go doesn't support getting coverage of multiple packages at the same time.

Is there a way to add a generic loop to this? Or perhaps we could add a bash script to call from .travis.yml?


cmbrose:
@lighthouse/owners tips for this would be much appreciated

nickmonad:
> [...] because Go doesn't support getting coverage of multiple packages at the same time.

wat.

Beyond that, it seems like we'd have to hard code in a list of packages to test and just do a bash for loop over that list, running the testing command for each. But, that seems really hacky and would incur slightly more overhead every time we add a new package.

nickmonad:
Could a python script handle that functionality? Then the travis file just has to call that?

cmbrose:
Off the top of my head, I don't know but I would imagine so.  I can explore that later today!

zach-taylor:
@cmbrose you could probably move all the current packages into one big package called 'lighthouse'. and then run the tests on that package. I think that's more of a standard thing to do just by looking at this repo: https://github.com/google/go-github

zach-taylor:
Otherwise here's a larger project with go code coverage: https://github.com/smira/aptly

zach-taylor:
@cmbrose did you try this exact configuration? https://github.com/mattn/goveralls#travis-ci

cmbrose:
@zach-taylor I would really be wary of turning everything into one big package! That'll create a huge mess I think.  The `aptly` example is just running a script essentially, so I think that that's probably the route we'll have to go (just not a Makefile like they use).

zach-taylor:
@cmbrose lol. I didn't mean one big package, just a main package with sub-packages.

chfogerty:
On mobile right now, but I can post that script I worked up earlier today. It just loops through directories/subdirectories and calls `go test`

cmbrose:
@chfogerty yeah, if you can either email that to me or commit it to this branch, that would be awesome.

cmbrose:
@lighthouse/owners sorry for the huge amount of worthless commits here, the only way I could really test was by pushing things up...

Anyways, as far as I can tell this is working! Coverage results are [here](https://coveralls.io/r/lighthouse/lighthouse) and update automatically after pushes.

Currently we're around 45% so there's room to grow.

zach-taylor:
Awesome! +1

zach-taylor:
@cmbrose oh, one thing. You can add this badge next to the others in the README so we have a quick a overview of the coverage: 

`[Coveralls](https://coveralls.io/r/lighthouse/lighthouse)`

It displays as 'unknown' right now because we haven't ran it on the master branch yet.

cmbrose:
@zach-taylor good call!  I'll add it right before this gets merged in then :)

cmbrose:
@lighthouse/owners the backend is starting to get a decent amount of testing utility files (for example, database setup) which is negatively impacting or code coverage (because we don't test testing utils...).  With the latest commit, any file ending in `_testing_utils.go` is ignored from coverage.

@chfogerty I know you really like the local HTML script too, so you can run this with as `./testcoverage.sh local` to generate the html instead.

PR #44, opened 2015-03-02
Bugfix - Docker payload

cmbrose:
Made payload map instead of string


cmbrose:
@ngmiller @theMagicalKarp any argument to merging this?

nickmonad:
looks good +1

PR #45, opened 2015-03-02
Auto aliasing

cmbrose:
This is the backend solution for the frontend sending Instance names as aliases.

Aliases for beacons and instances are now automatically generated when they are added.  The `/beacons/create` endpoint now has a required `Alias` field, and instance aliases are generated as `<Beacon Alias>.<Instance Name>`.

These new aliases are now included in the `/beacons/list/{Beacon}` endpoint as

``` json
[
    {
        "Alias" : "The.Alias",
        "// Previously existing fields..."
    }
]
```

For backend use, aliases are now keyed by the address, instead of the alias.  Aliases no longer need to be unique, but instances _must_ be.


cmbrose:
@ngmiller will this work for you to fix this?

nickmonad:
Looks good! Will open a client PR for the required alias key on creation. Quick question though: Is it possible to catch these postgres errors for duplicate keys and return a friendlier message? Something along the lines of "a beacon with the given address already exists" (possibly with a 400 status code?)

screen shot 2015-03-02 at 7 49 10 pm

cmbrose:
@ngmiller I'm looking into that.  The problem is `lib/pq` isn't returning normal errors, it's actually giving us the type here, http://godoc.org/github.com/lib/pq#Error, but it's hidden under polymorphism so it's hard for me to extract the exact error.  Without making this postgres specific that is.

**EDIT:** I have an idea for fixing this, but it means adding error transforms to our postgres package.  Easy, but out of scope for this PR.

nickmonad:
+1

robertsheehy-wf:
+1

PR #48, opened 2015-03-04
Better database errors

cmbrose:
## What

Extended the postgres package to capture pq Errors and turn them into database errors
## Why

Allows database users to not have to care about what driver is being used for error checking.  All errors should either be SQL or database types.
## How

postgres catches errors and checks if it is a pq.Error.  If it is, it simply swaps out that error for a better (more user friendly) one.

### DIFF
--- a/databases/postgres/postgres.go
+++ b/databases/postgres/postgres.go
@@ -21,21 +21,44 @@ import (
 
     "database/sql"
 
-    _ "github.com/lib/pq"
+    "github.com/lib/pq"
 
     "github.com/lighthouse/lighthouse/logging"
+
+    "github.com/lighthouse/lighthouse/databases"
 )
 
-var connection *sql.DB
+type postgresConnection struct {
+    sql.DB
+}
 
-func Connection() *sql.DB {
+var connection *postgresConnection
+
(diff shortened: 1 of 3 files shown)


cmbrose:
@ngmiller this should fix your duplicate alias error code issue

databases/postgres/postgres.go
```diff
@@ -21,21 +21,55 @@ import (
...
+        pqErr, ok = err.(*pq.Error)
+    } 
+
+    if !ok {
+        return nil
+    }
+
+    switch pqErr.Code {
+    case "23505": 
+        return databases.DuplicateKeyError
```
cmbrose, line 66: @ngmiller @zach-taylor I'm only catching this one specific error currently, so if you've seen other error text that looks like `pq: ...`, let me know and I'll add it here.

cmbrose:
@ngmiller @zach-taylor were there any other errors you've seen coming up that should get patched?  Otherwise I'd like to get this guy through.

nickmonad:
Nope I think that's the only one I've seen thus far. Can always add more later. +1

chfogerty:
+1

PR #49, opened 2015-03-05
Ensure content type is application/json

nickmonad:
When testing container creates, I found that lighthouse is either stripping the `Content-Type` header or simply not setting it. This results in a 500 response code from the targeted docker instance.

### DIFF
--- a/handlers/docker/docker.go
+++ b/handlers/docker/docker.go
@@ -71,6 +71,8 @@ func DockerRequestHandler(w http.ResponseWriter, info handlers.HandlerInfo) *han
         req.Header.Set(beacons.HEADER_TOKEN_KEY, token)
     }
 
+    // Ensure Content-Type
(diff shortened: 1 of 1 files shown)


nickmonad:
@lighthouse/owners Check it, yo.
@theMagicalKarp Should we do something similar for beacon? (i.e. first checking then defaulting to application/json)

robertsheehy-wf:
+1 nice, I was just about complain about varying content types, but you fixed it. ;)

robertsheehy-wf:
@ngmiller This is almost exactly the same as beacon. I don't think think defaulting to `application/json` will matter too much.  In any case the client can specify the content type and override the default.

nickmonad:
@robertsheehy-wf Agreed

chfogerty:
+1

nickmonad:
Verbal +1 from @cmbrose

PR #50, opened 2015-03-09
Stream responses from beacon

theMagicalKarp:
This ensures that docker responses are streamed back from lighthouse.  This allows for Oboe to parse the streams and do status updates on image pulls and logs.  This change should be pretty begin in regards to existing docker calls since it's all the same to angular `$http`.

This is related to https://github.com/lighthouse/beacon/pull/12


theMagicalKarp:
@lighthouse/owners

handlers/docker/docker.go
```diff
@@ -86,13 +85,18 @@ func DockerRequestHandler(w http.ResponseWriter, info handlers.HandlerInfo) *han
...
+    w.WriteHeader(resp.StatusCode)
+    var bodyBuffer = make([]byte, 16)
+
+    for {
+        n, err := resp.Body.Read(bodyBuffer)
+        w.Write(bodyBuffer[:n])
+
+        if err != nil {
+            break
+        }
```
cmbrose, line 110: Should the break be flipped with the Write() call?

robertsheehy-wf, line 110: This is because the error returned is usually a signal that notifies us we have reached the end of a stream.  What's left in `bodyBuffer` is the last few bytes in the stream that were read before the `EOF` was reached.  If we didn't write that last buffer bytes our message would be truncated.

cmbrose, line 110: That makes sense.  What you have is fine then, but you could do something like

``` go
for err == nil {
    n, err := resp.Body.Read(bodyBuffer)
    if err != io.EOF {
        break
    }
    w.Write(bodyBuffer[:n])
}
```

handlers/docker/docker.go
```diff
@@ -86,13 +88,21 @@ func DockerRequestHandler(w http.ResponseWriter, info handlers.HandlerInfo) *han
...
 
-    body, err := ioutil.ReadAll(resp.Body)
-    if err != nil {
-        return &handlers.HandlerError{500, "control", "Failed reading response body"}
+    w.WriteHeader(resp.StatusCode)
+    var bodyBuffer = make([]byte, 16)
+
+    for {
+        n, err := resp.Body.Read(bodyBuffer)
+        w.Write(bodyBuffer[:n])
```
robertsheehy-wf, line 103: @cmbrose
The document seems to support the idea of always processing the bytes before handling the error.  As long as we use `n` to slice the byte array I think we should be fine.

> Callers should always process the n > 0 bytes returned before considering the error err. Doing so correctly handles I/O errors that happen after reading some bytes and also both of the allowed EOF behaviors.

http://golang.org/pkg/io/#Reader

I do however log any unexpected exception now.

nickmonad:
@theMagicalKarp Could you merge master into this? There some auth updates and other things needed to match the state of client/add-images

robertsheehy-wf:
@ngmiller done

nickmonad:
+1

chfogerty:
+1

cmbrose:
+1

PR #51, opened 2015-03-09
User templates on index.html

cmbrose:
This supports the auth refresh issue on the client by populating the user email template field in `index.html`.

PR #52, opened 2015-03-12
Backend error standardization

cmbrose:
**NOTE** this replaces https://github.com/lighthouse/lighthouse/pull/51
## What

Standardize all of Lighthouse's errors into a JSON scheme.  All codes >= 400 will also have JSON of the form:

``` json
{
    "Cause" : "<Package the error came from>",
    "Message" : "<Reason for the error>"
}
```
## Why

In reference to https://github.com/lighthouse/lighthouse-client/pull/34, the frontend handles all errors through one function.  This means errors need to always be of the same form.
## How

To ensure that errors are always the same, all error writes should now be processed through the `WriteError()` function in the `handlers` package:

``` go
func WriteError(w http.ResponseWriter, code int, cause, message string)
```

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -95,7 +95,7 @@ func AuthMiddleware(h http.Handler, ignorePaths []string) http.Handler {
             return
         }
 
-        if strings.Contains("/api", r.URL.Path) {
+        if strings.HasPrefix(r.URL.Path, "/api") {
             w.WriteHeader(401)
         } else {
             http.Redirect(w, r, "/", http.StatusFound)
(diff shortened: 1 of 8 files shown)


cmbrose:
@chfogerty I'm assuming that you return HandlerErrors from your `container` handlers, so this shouldn't affect https://github.com/lighthouse/lighthouse/pull/41.

nickmonad:
Hitting some invalid JSON with an incorrect password attempt `{"Cause":"auth","Message":"email or password incorrect"}false` Looks like it's originating from here? https://github.com/lighthouse/lighthouse/blob/error-standardize/auth/auth.go#L147

cmbrose:
@ngmiller good catch - I had to update every place that we write responses for this PR, so it's very possible that I missed a couple ;)  Fixed this one though.

nickmonad:
@cmbrose No worries! Looks good to me +1

chfogerty:
@cmbrose Correct.

PR #54, opened 2015-03-20
Added reload and driver flags

cmbrose:
## What

Added flags to lighthouse.go to control whether or not databases are reset at load and what driver to use.
## Why

Because blowing up the users database every time the program starts is probably a bad idea.
## How

Lighthouse can now be run with the flags 
- `-databases-reload` - add this to start _all_ databases from config data.
- `-databases-driver <DRIVER>` - specify what driver to use for databases (currently the only supported value is `postgres`).  Default value is `postgres`.


cmbrose:
@lighthouse/owners for your review.  Also completely open to ideas of how to do the drivers (or any of this really) in a cleaner way.

nickmonad:
+1

robertsheehy-wf:
+1

PR #55, opened 2015-03-23
Fixed 'SELECT *' issue

cmbrose:


cmbrose:
Verbal +1 from @chfogerty

PR #57, opened 2015-03-28
Database drivers

cmbrose:
## What

A major overhaul of the databases package to lay groundwork for additional drivers.
- Removed all non-schema methods - all tables **must** be schema-ed
- Solved the `INSERT ... RETURNING` compatibility issue
- Solved the issue of different database types using different syntaxes
## How

>  Solved the `INSERT ... RETURNING` compatibility issue
- Created a `LockingTable` (really just an option field in the existing `Table` type.
  - The table has a `sync.Mutex` which is used to lock the table
  - Currently only `INSERT`s lock the table
- Broke the `Insert(values, returning)` method into `Insert( ... )` and `InsertReturn( ... )`
  - `Insert( ... )` locks the table, runs an `INSERT`, and unlocks
  - `InsertReturn( ... )` locks the table, runs an `INSERT` followed by a `SELECT`), and unlocks
  - All tables can call `Insert()`, only LockingTables can call `InsertReturn()`

>  Solved the issue of different database types using different syntaxes
- Added the `databases.Compiler` type
  - Knows the syntax of one database type and, given parameters, creates Execs and Queries for that type
  - Each database will probably need its own version (there may be small overlap*)
- Added a `Compiler()` method to `DBInterface` which gets a compiler for that database

``` go
db := postgres.Connection()
comp := db.Compiler(mySchema)
query, args := comp.CompileSelect( ... )
```
## Why

Supporting new databases other than postgres.  The syntax differences alone were enough to make adding new types impossible in our current structure. 
## Testing

Tests for the database package appear weaker, but are actually just as strong.  This is because the database relies on a compiler to function properly, but like usual we can't rely on one specific type of compiler.  As mock compiler is used in the tests which creates queries that are simply the command (`INSERT`, `DELETE`, etc).  Because `sqlmock` is handling all our return values, these simple commands don't affect any actual behaviors.

Tests are added for the postgres package to ensure that it creates the right queries for the various commands.

_*_ the postgres compiler is now our largest source file sooooo... yeah.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -39,7 +39,7 @@ var SECRET_HASH_KEY string
 
 func Init(reload bool) {
     if users == nil { // defined in users.go
-        users = databases.NewSchemaTable(databases.DefaultConnection(), "users", schema)
+        users = databases.NewTable(databases.DefaultConnection(), "users", schema)
     }
 
     config := LoadAuthConfig()
--- a/auth/user_test.go
+++ b/auth/user_test.go
@@ -48,7 +48,7 @@ func addUsers(list ...User) {
 	        "Password" : user.Password,
             "AuthLevel" : user.AuthLevel,
 	        "Permissions" : user.Permissions,
-	    }, "")
+	    })
 	}
 }
 
@@ -81,7 +81,7 @@ func Test_CreateUser(t *testing.T) {
     }
 
     var actual User
-    users.SelectRow(nil, nil, &actual)
+    users.SelectRow(nil, nil, nil, &actual)
 
 	assert.Equal(t, keyUser, actual)
 }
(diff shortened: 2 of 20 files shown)


cmbrose:
@lighthouse/owners for your review.  I'd like to make sure that I didn't break anything before I go too crazy with the new drivers.  As far as I can tell this is working perfectly, but more eyes is always good.

Also, I know this affects a lot a files, but most of them are just method renames.  The only real changes here are in databases.go and postgres.go.

cmbrose:
@lighthouse/owners in light of our discussion today, I'm prioritizing rollbacks and history over drivers.  I'm going to build the rollback PR off this, but I am cutting off development of this branch.

If time permits I might make a new PR to add actual drivers.  At this point we can just say that it's easy to do.

nickmonad:
+1

PR #58, opened 2015-03-30
Read session key from file

robertsheehy-wf:
Read the session key from a tmp file to help persist user sessions.  Currently if lighthouse restarts the user is forced to re-auth. This fixes that.

### DIFF
--- a/session/session.go
+++ b/session/session.go
@@ -15,13 +15,39 @@
 package session
 
 import (
+    "os"
+    "io/ioutil"
     "net/http"
 
     "github.com/gorilla/sessions"
     "github.com/gorilla/securecookie"
 )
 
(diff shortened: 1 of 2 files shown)


nickmonad:
What's the policy for session expiration? Should we compliment this with a cron script to clear out the temp files?

robertsheehy-wf:
Currently sessions do not have an expiration date.
Also if you want we can use [github.com/robfig/cron](https://godoc.org/github.com/robfig/cron) to schedule key regeneration.  Sounds like a cool idea but I'm not sure how practical it is.

nickmonad:
Generally speaking, it's probably not a good idea to have a session never expire. I'd rather force a relogin if lighthouse restarts.

@lighthouse/owners Thoughts?

PR #59, opened 2015-03-30
Test beacon validity on create

cmbrose:
If there is an error getting the VMs from a beacon in `/beacons/create` then the beacon (and alias) are not added.  An error is returned, just like it was originally.

### DIFF
--- a/beacons/aliases/aliases.go
+++ b/beacons/aliases/aliases.go
@@ -57,6 +57,11 @@ func AddAlias(alias, address string) error {
     return err
 }
 
+func RemoveAlias(address string) error {
+    where := databases.Filter{"Address" : address}
+    return aliases.DeleteRowsSchema(where)
+}
+
 func UpdateAlias(alias, address string) error {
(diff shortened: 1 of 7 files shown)


nickmonad:
Can we dial back the timeouts?

```
2015/03/30 15:25:47   42.29058ms GET /api/v0.2/beacons/list/192.168.59.103:5000
2015/03/30 15:26:34 30.131982895s POST /api/v0.2/beacons/create
2015/03/30 15:28:51 30.086733297s POST /api/v0.2/beacons/create
```

cmbrose:
@ngmiller definitely!  What seems reasonable? 3 seconds?

nickmonad:
Sounds good for now :)

cmbrose:
@ngmiller can you check that this the timeout works now? On my machine I get a connection refusal almost immediately (it doesn't timeout).

nickmonad:
@cmbrose Works for me! +1

